PR #5, opened 2016-01-31
Add select-fields to xml actions.

chunlinyao:
In some situation, The <select-field /> is changed by user input. with <select-fields /> I can construct a list and pass it to <entity-find />. A select-field.@ignore attribute can also work, But <select-fields /> is easy to implement.


jonesde:
The general need is there, but I'd rather use an approach like the EntityFind.orderBy() method, supporting comma separated field names. That is done in commit #2b33809.

PR #7, opened 2016-02-02
Improved the getRuntime task in order to get the specified runtime

zhangshifeng:


jonesde:
Looks good. This is usually called automatically, but yes handy to be able to get the specific runtime you want.

PR #8, opened 2016-02-02
Source compatibility

zhangshifeng:


jonesde:
What is the reason for these changes? The sourceCompatibility is useful for creating distributions and such, and documents that Java 1.7 or later is what is supported. If you are trying to build and run on JDK 1.6 it won't work, there are a few libraries that require Java 1.7 or later.

zhangshifeng:
I have some legacy systems run on the JDK6, so wanna test about source compatibility(just interested).

I'm sorry for didn't test enough(only tested based on JDK8 and set sourceCompatibility=1.6).

Now i test on JDK6, i get some error as you said. Sorry again :-)

PR #11, opened 2016-02-04
When using dependencyUpdates don't show pre-releases

samhamilton:
A quick change to help the report for dependencyUpdates be more useful by not showing pre-release software.


jonesde:
That will be helpful, thanks Sam.

PR #13, opened 2016-02-05
Improve setSequencedIdSecondary performance

chunlinyao:
I have a slow service when create 5000 records. logProfilingDetail() shown many time is used by setSequencedIdSecondary. Though use secondary sequence id for large data set is a design mistake. This change make create 5000 records five times fast.

I have another version use plain SQL  and is more faster. Fetch one line `order by char_length(seqId) desc, seqId desc`, But It seems plain SQL method is over optimized.


jonesde:
Thank you Yao, that's a good change. In general if you are doing large sets of sequenced ID records it's better to use a primary sequenced ID with a large bank size set on the entity (or in the bankSize parameter to the EntityFacade.sequencedIdPrimary() method). This will run MUCH faster as there is, for example, only one DB find and one create/update per bank (i.e. per 100 or 1000 or whatever you set the bank size to).

PR #19, opened 2016-02-11
minor fixs

chunlinyao:


jonesde:
I just committed a huge set of changes to use a new class (MNode) for XML handling instead of groovy Node. As part of those changes I noticed this issue and a fix is in place for it (though I haven't tested it on Atomikos yet).

PR #20, opened 2016-02-13
2016-02-12

sbessire:


jonesde:
It's an interesting idea wrapping DataSource and XAResource, would be nice to avoid the separate thread. For now it seems to be a workable solution. If you're okay with submitting those tests please do, would be nice to have some quick tests in framework to do this (I've been using the mantle-usl tests which do it in a few places, but they take a bit longer to run).

BTW, I'm considering changing how getConnection is managed to reuse connections more without going through Bitronix every time. Bitronix is pretty fast, way faster than Atomikos, but in profiling I still see some slightly inefficient code that could be improved and as I improve other things Bitronix stuff is rising more and more in the hot spots list.

PR #25, opened 2016-02-19
Support expression in complex-alias

chunlinyao:
This change allow use sql expression in complex-alias. Which discussed in LinkedIn https://www.linkedin.com/groups/4640689/4640689-6083159241856737285 

Because complex-alias allow nested, so I added `complex-alias.@expression` other  than `alias.@expression`. Here is some usage example I tested.

A simple expression with constant value

        <alias name="test1" type="number-decimal">
            <complex-alias operator="*">
                <complex-alias-field entity-alias="AST" field="quantityOnHandTotal" />
                <complex-alias operator="" expression="100" />
            </complex-alias>
        </alias>

A sample with complex function

        <alias name="test2" type="number-integer">
            <complex-alias operator="" expression="EXTRACT(YEAR FROM ">
                <complex-alias-field entity-alias="AST" field="lastUpdatedStamp" />
                <complex-alias operator="" expression=")" />
            </complex-alias>
        </alias>

A sample use a whole expression

        <alias name="test3" type="number-decimal">
            <complex-alias operator="/" expression="FLOOR(AST.QUANTITY_ON_HAND_TOTAL / 1000)" />
        </alias>

A sample use simple function

        <alias name="test3_alt" type="number-decimal">
            <complex-alias operator="/" expression="FLOOR" >
                <complex-alias-field entity-alias="AST" field="quantityOnHandTotal" />
                <complex-alias operator="" expression="1000" />
            </complex-alias>
        </alias>

A sample with case when.

        <alias name="test4" type="text-short">
            <complex-alias operator="">
                <complex-alias operator="/" expression="CASE WHEN" >
                    <complex-alias-field entity-alias="AST" field="quantityOnHandTotal" />
                    <complex-alias operator="" expression="1000" />
                </complex-alias>
                <complex-alias operator="" expression="BETWEEN 0 AND 3 THEN 'LOW' END" />
            </complex-alias>
        </alias>

### DIFF
--- a/framework/src/main/groovy/org/moqui/impl/entity/EntityDefinition.groovy
+++ b/framework/src/main/groovy/org/moqui/impl/entity/EntityDefinition.groovy
@@ -738,6 +738,11 @@ public class EntityDefinition {
     }
 
     protected void buildComplexAliasName(MNode parentNode, String operator, StringBuilder colNameBuilder) {
+        //TODO expand expression against current context
+        colNameBuilder.append(parentNode.attribute('expression') ?: '')
+        if (parentNode.children.isEmpty()) {
+            return
+        }
         colNameBuilder.append('(')
         boolean isFirst = true
         for (MNode childNode in parentNode.children) {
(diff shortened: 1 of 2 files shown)


jonesde:
This is an okay approach, seems to handle quite a few different things. I would still like to implement some of the designs shown in the commented InvoiceAgingSummary view-entity in the AccountingAccountViewEntities.xml file in mantle-usl, but this is worth having now so thank you!

Issue #42, opened 2016-04-03
StackOverflowError when a pre-validate seca rule call a service with a "parameters" in-param


chunlinyao:
Parameters contains key of 'context', context contains 'parameters' key,  countArtifactHit method concat parameters and throw StackOverflowError

jonesde:
In other words you declared an in-parameter called 'parameters' and it is a Map with an entry that has a key of 'context' and the value is the ec.context Map (or ContextStack rather)?

Yeah, don't do that. Certain names are not valid, or are reserved words, for service parameters including 'parameters', 'results', and 'context'. The Service Facade adds these to the context on its own to make them available to a service, and does special handling with them.

chunlinyao:
Yes, you are right. The context is in parameters because when a transition calling service, it pass the whole context as in-map (needless keys will be removed after `convertValidateCleanParameters`). Service engine add 'parameters' into context before call a seca rule, because some seca rule will modify the parameters of service. It is straightforward to define a in-parameter as 'parameters'. Even I change the parameter name declared in in-parameter to others, and use a `<set field="others" from="parameters"/>` in actions of seca rule. The problem will not resolve. Because it's nothing about parameter name, It is really a parameter (Map) has indirect reference to itself. I think the bug is caused by `context` become a real key recently.

chunlinyao:
A workaround like move parameters modification logic into actions of seca rule seems work.

jonesde:
Do you have an easy way to reproduce this? Some code snippets would be helpful (like the SECA rule you are using).

Also, what do you mean by "move parameters modification logic into actions of seca rule"?

chunlinyao:
I will try to create a minimal test code for it.  "move parameters modification logic into actions of seca rule” means write the parameters modification logic in <actions> tag of seca file (<actions><service-call …/><set field=“parameters.foo” from=“bar” /></action>), that not require to pass the parameters as a in-parameter into the service be called by seca rule.

chunlinyao:
This is the test file https://gist.github.com/chunlinyao/cb71bc98ae7da8f9e4c294368f509ac6

Can reproduce before e7f5234e3bef24978b7eea99e0f712fa47304dfe

chunlinyao:
After e7f5234 this can not be reproduced, I think it's ok to close it.

PR #43, opened 2016-04-04
bugfix #42 prevent recursive toString call on ContextStack.

chunlinyao:


jonesde:
Doesn't the existing code already work to prevent this? One issue with the approach in this patch (using a field on a ContextStack instance that is set when printing) is that it is not thread-safe and while that's normally okay (most of the ExecutionContext is not thread safe, it is not meant to be) this would also have issues if another ContextStack instance was in the context, which is done in a few places in the framework! One example is the combined parameters maps in WebFacadeImpl. In other words the ContextStack class is not used only for ec.context.

chunlinyao:
Existing code only prevent a ContextStack in ContextStack directly. This issue is ContextStack -> HashMap -> ContextStack. When call toString every ContextStack instance in the context will only output once, so I don't see any issues with another ContextStack instance in the context.

jonesde:
I see, yes, it would print each ContextStack once if a different object is nested somewhere. On the thread safe thing it should be fine, I don't think it's used that way anywhere, and worst case scenario is a ContextStack doesn't print in toString() because it's printing in another thread. I'll merge this.

PR #46, opened 2016-04-20
Added minxed=true in parameter element in service-definition.xsd so t…

shendepu:
…hat description of parameter is allowed to added without exception of 'element parameter cannot have character [children], because type's content type is element-only'

### DIFF
--- a/framework/xsd/service-definition-1.6.xsd
+++ b/framework/xsd/service-definition-1.6.xsd
@@ -215,7 +215,7 @@ along with this software (see the LICENSE.md file). If not, see
 
     <xs:element name="ParameterValidations" abstract="true"/>
     <xs:element name="parameter">
-        <xs:complexType>
+        <xs:complexType mixed="true">
             <xs:sequence>
                 <xs:element minOccurs="0" ref="description"/>
                 <xs:element minOccurs="0" maxOccurs="unbounded" ref="subtype"/>


jonesde:
The parameter element should not have character children, just elements. It is the description element under the parameter that should have character children.

Am I missing something?

shendepu:
Sorry, it should be mixed=true in "in-parameters", so as "out-parameters". 

I did the test again, it is so strange. I keep the xsd unchanged, then adding description for parameter in "in-parameter" pop up the exception, but not in "out-parameters" 
image
When set mixed=true for "in-paraeters"

    <xs:element name="in-parameters">
        <xs:complexType mixed="true">
            <xs:sequence>
                <xs:choice minOccurs="0" maxOccurs="unbounded">
                    <xs:element ref="auto-parameters"/>
                    <xs:element ref="parameter"/>
                </xs:choice>
            </xs:sequence>
        </xs:complexType>
    </xs:element>

then the error in Intellij IDEA is gone. 

I google the error, solution is said to add mixed=true. And if this is ok to make the annoying error gone by this way, I will update it to right place and update pull request.

jonesde:
That is strange... also notice the characters under in-parameters are Chinese while under out-parameters they are latin, maybe IDEA doesn't support multi-byte characters so well?

Oh wait, one more thing: you have an extra greater-than (>) after the close parameter tag... it's probably that.

shendepu:
I am so stupid... thanks pointing out the extra (>).  the error is gone now. 

This is pull request is useless.

PR #48, opened 2016-04-21
Dialog icons

jenshp:
Add a button-icon attribute to container-dialog and dynamic-dialog.
This goes together with https://github.com/moqui/moqui-runtime/pull/25 that uses these attributes.


jonesde:
I considered this a long time ago when implementing these and decided it wasn't a good idea, the main reason being to have a consistent icon so users know when a button is for a popup window (i.e. that it will always show a popup/modal window as opposed to doing something or going somewhere else).

In other words my initial reaction is no, we shouldn't even allow this... just opens things up for a lesser user experience.

Thoughts?

PR #60, opened 2016-06-01
tenantAllowOverride=N should only reject change to other tenants not the tenant bind to the hostname.

chunlinyao:
I created a tenant and a `TenantHostDefault`, But I can open the login screen using that hostname. The following error occurs `Tenant override is not allowed for host [foobar]`. I think `tenantAllowOverride=N` should only prevent change to other tenantId. `initWebFacade` method change tenantId from  `DEFAULT` to the tenantId of this hostname.


jonesde:
Looks good to me, merging.

PR #62, opened 2016-06-03
Localized field in view

jenshp:
When a field with enable-localization=true is an alias in a view, it would be expected to be localized in the view as well. These changes do exactly this, provided that the PK of the member entity with the localized field is also included in the view.
As I'm not quite familiarized with the internal workings of moqui-framework the changes could surely be improved, so if there are suggestions, they are welcome.

### DIFF
--- a/framework/src/main/groovy/org/moqui/impl/entity/EntityDefinition.groovy
+++ b/framework/src/main/groovy/org/moqui/impl/entity/EntityDefinition.groovy
@@ -202,11 +202,7 @@ public class EntityDefinition {
                 if (fieldNode == null) throw new EntityException("In view-entity [${internalEntityName}] alias [${aliasNode.attribute("name")}] referred to field [${fieldName}] that does not exist on entity [${memberEd.internalEntityName}].")
                 if (!aliasNode.attribute("type")) aliasNode.attributes.put("type", fieldNode.attribute("type"))
                 if (fieldNode.attribute("is-pk") == "true") aliasNode.attributes.put("is-pk", "true")
-                if (fieldNode.attribute("enable-localization") == "true") {
-                    aliasNode.attributes.put("enable-localization", "true")
-                    aliasNode.attributes.put("view-original-entityName", memberEd.fullEntityName)
-                    aliasNode.attributes.put("view-original-fieldName", fieldName)
-                }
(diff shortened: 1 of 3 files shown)


jonesde:
The concept looks good Jens, is a helpful feature. One small thing and one big thing...

Small: please use 4 spaces instead of tabs for indentation.

Big: the information about the entity and field the aliased field came from is available without adding internal attributes for them (and additional fields on the FieldInfo class), better to get the field name as is done else where using alias.@field ?: alias.@name and look up the entity name based on the alias.@entity-alias attribute and the corresponding member-entity.

jenshp:
I just added changes that avoid the addicional internal attributes. Had to add a method to the EntityDefinition class that gets the member-entity name from the member-alias.

jenshp:
There was a bug in this branch that surfaced when the primary key from the member entity had a different name in the alias.

jonesde:
Thank you Jens, looks good overall. I'll merge this and try it out.

jonesde:
Quick note on this: after applying this patch I'm seeing tons of warnings with Mantle/etc in place like: 

Not all primary-key fields in view-entity [moqui.internal.DynamicView] for member-entity [moqui.basic.Enumeration], skipping cache reverse-association, and note that if this record is updated the cache won't automatically clear; pkFieldNames=[enumId]; partial mePkFieldToAliasNameMap=[:]

I haven't yet looked into why this is happening, but appears to be caused by these changes.

jenshp:
Are you seeing this during testing?
How are you doing the testing btw? When I attempt to start the tests using gradle I only get "UP-TO-DATE" for mantle-usl:test and others. Am I missing something?

jenshp:
Got them now, I was using a link to the runtime instead of a directory within the framework repository, which was confusing gradle and it did not find the directories to execute the tests.

jonesde:
I have a temporary fix for this going in soon, basically don't try view-entity localization if it is a dynamic view entity. We may be able to get it working, but for now it avoids the issue.

jonesde:
With gradle if the tests were already compiled (no changes, no clean) then it will show UP-TO-DATE and won't run the tests again. You can force a test run with a clean then test. In some of my build.gradle files for components I have even added a 'test.dependsOn clean' line so that tests always run.

PR #68, opened 2016-06-20
Translate transition name and description.

jenshp:
Allow automatic transition links to be translated.

### DIFF
--- a/framework/data/CommonL10nData.xml
+++ b/framework/data/CommonL10nData.xml
@@ -75,7 +75,7 @@ along with this software (see the LICENSE.md file). If not, see
 
     <!-- Message templates for formatting common sets of entity fields (generally for a view-entity) -->
     <moqui.basic.LocalizedMessage locale="default" original="StatusTransitionNameTemplate"
-                                  localized="${transitionName} (${description})"/>
+                                  localized="${ec.l10n.localize(transitionName)} (${ec.l10n.localize(description)})"/>
     <moqui.basic.LocalizedMessage locale="default" original="UomNameTemplate"
                                   localized="${description?:''} (${abbreviation?:uomId})"/>
     <moqui.basic.LocalizedMessage locale="default" original="UserGroupNameTemplate"


chunlinyao:
Seems ok to me.

jonesde:
It's a little funky doing double localization... that might be a good thing to keep an eye out for as a not recommended practice. In this case both transitionName (StatusFlowTransition) and description (StatusItem) are coming from entity fields that can be localized, and that is better than doing separate localizations (for context, reuse, etc, etc). The StatusItem.description field already has enable-localization=true, and I'm adding that to StatusFlowTransition.transitionName. That is done in commit #125547e.

There might be other cases where double localization would be useful, but as a generality it strikes me as something cumbersome and error prone.

jenshp:
That was actually my first attempt, but since StatusFlowtransition has 3 PK
fields and LocalizedEntityField works if the entity with the localized
field has only one PK, or has this changed?

Regards,

El lun., 20 jun. 2016 a las 21:35, David E. Jones (<notifications@github.com>)
escribió:

> Closed #68 <https://github.com/moqui/moqui-framework/pull/68>.
>
> —
> You are receiving this because you authored the thread.
> Reply to this email directly, view it on GitHub
> <https://github.com/moqui/moqui-framework/pull/68#event-698446431>, or mute
> the thread
> <https://github.com/notifications/unsubscribe/AMnAHEKk4-OM6IqfWDfNwzN_WyVtghZQks5qNz_EgaJpZM4I6F2x>
> .
>
-- 

Jens.

jonesde:
Yes, that's an important point... if an entity has more than one PK field then it won't do a query on LocalizedEntityField, but if enable-localization=true for a field then it will do a query for the original (the field's DB value) in LocalizedMessage. In other words, as long as that flag is set it shouldn't be necessary to explicitly localize (by ec.l10n.localize) a value from an entity field.

Let me know if that doesn't work. BTW, you should be able to use LocalizedMessage for the StatusItem field too so that the localized messages are handled more consistently in your data.

jenshp:
It worked fine, thanks!

El lun., 20 jun. 2016 a las 22:38, David E. Jones (<notifications@github.com>)
escribió:

> Yes, that's an important point... if an entity has more than one PK field
> then it won't do a query on LocalizedEntityField, but if
> enable-localization=true for a field then it will do a query for the
> original (the field's DB value) in LocalizedMessage. In other words, as
> long as that flag is set it shouldn't be necessary to explicitly localize
> (by ec.l10n.localize) a value from an entity field.
>
> Let me know if that doesn't work. BTW, you should be able to use
> LocalizedMessage for the StatusItem field too so that the localized
> messages are handled more consistently in your data.
>
> —
> You are receiving this because you authored the thread.
> Reply to this email directly, view it on GitHub
> <https://github.com/moqui/moqui-framework/pull/68#issuecomment-227325517>,
> or mute the thread
> <https://github.com/notifications/unsubscribe/AMnAHBjxV2jLP_H7hJRtXtzkcyacdF9Nks5qN066gaJpZM4I6F2x>
> .
>
-- 

Jens.

PR #70, opened 2016-06-22
Format fraction digits according to currency

jenshp:
Use the default fractionDigits number to format currencies. This way we do not specify them hardwired in the screens, but let the framework figure out how to format.

If this is OK, it would also be necessary to provide a way to format the text-line tags containing currency values, currently explicitly formatted in screens using "#0.00" that also should depend on the currency of the value being displayed.
As they should not contain the symbol prefix nor thousands separators to facilitate editing, they should be formatted using other methods or adding a parameter indicating whether to use thousands separators and prefixes.

### DIFF
--- a/framework/src/main/groovy/org/moqui/impl/context/L10nFacadeImpl.java
+++ b/framework/src/main/groovy/org/moqui/impl/context/L10nFacadeImpl.java
@@ -98,6 +98,11 @@ public class L10nFacadeImpl implements L10nFacade {
     }
 
     @Override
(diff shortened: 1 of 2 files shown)


jonesde:
This looks fine Jens, I agree it is a generally good approach. Still having explicit digits is important as in some cases we want to display more than the default currency digits, i.e. order/invoice/etc items typically have one more digit than the default currency digits. We may need to do something like an additional digits parameter beyond the default... but we can start with this.

PR #83, opened 2016-07-29
Throws AuthenticationRequiredException instead of ArtifactAuthorizati…

shendepu:
…onException when ArtifactExecutionFacadeImpl.isPermitted=false and user is not logged in


jonesde:
NOTE: this is related to the discussion at:

https://www.linkedin.com/groups/4640689/4640689-6164327774686371841

jonesde:
I don't want to make this change in pushInternal(), which should only be responsible for checking authorization, not authentication. Authorization should be checked in various places higher up. I'll continue the conversation about this in the LinkedIn thread.

PR #89, opened 2016-08-17
Suggest a workaround for bug in xalanj (XALANJ-2195)

chunlinyao:


jonesde:
Looks good to me, hopefully we won't need it forever but shouldn't add much overhead anyway.

PR #98, opened 2016-08-31
Fixed issue of should lock parent resource in DbResource instead of D…

shendepu:
…bResourceFile


jonesde:
Thank you Jimmy, good catch! Apologies that I rushed this and didn't review or test it thoroughly, admittedly it is partly because I knew you would be testing and reviewing it.

PR #113, opened 2016-09-14
Fixed response Stream closed issue when sendJsonError in handleServic…

shendepu:
…eRestCall.


jonesde:
Thank you Jimmy, that is a good fix.

PR #115, opened 2016-09-14
Set request attribute moqui.request.loginKeyValid=false if loginKey i…

shendepu:
…s not valid, then handleServiceRestCall can send 401 error response


jonesde:
Implemented this a bit differently, handle more general login failures. See commit #935a97f.

PR #130, opened 2016-10-07
Added notNull property in FieldInfo

shendepu:


shendepu:
The entity.field.@not-null defines whether a field is allowed null, but this info is not stored in FieldInfo class. It does not have impact on db since EntityDbMeta checks it.

However, when I construct GraphQL Object type and Field based on EntityDefinition, it needs to know whether a field allows null or not.

jonesde:
That class is a low-level internal class used for fast access at runtime to various entity field attributes and other settings. It is not meant to represent all field attributes or options, including (so far) the @not-null attribute. You can get it from your code using the FieldNode.fieldNode.attribute() method, along with any others you need. You can also get the field MNode object directly from the EntityDefinition to avoid getting too deep in the low-level stuff that tends to change more over time (for optimizations and such... all of this is pretty well optimized now and may not change more, but this is the sort of low-level thing that tends to change).

PR #132, opened 2016-10-12
Fixed field-map issue in parent and related relationship of Enumerati…

shendepu:
…on entity


shendepu:
It seems when the key-map.@related is missing, it should be set as primary key from related-entity definition as . And if this is the desired behavior, then this PR is obsolete, but a bug should be raised. 

Current the Master-Detail.relInfo.@keyMap holds ["parentEnumId": "parentEnumId"] for parent relationship in Enumeration, which should be ["parentEnumId": "enumId"]

shendepu:
After look at the code, the behavior is let relationship.@related = relationship.@fieldname if relationship.@related is missing. Then this PR seems valid.

jonesde:
I made a change to the EntityDefinition method that creates the relationship key maps to resolve this issue. It has come up a few times over the years and the new code is less error prone for cases like this. The old approach was, even for single PK entities, to use a related entity field name if there is one that matches. In the new code if there is only one PK field and it is a type one or one-nofk relationship then it always uses that PK field. With that change the Enumeration parent relationship now has the correct key maps.

PR #136, opened 2016-10-13
Rest client https

jenshp:
When using RestClient to connect to another server using an https URL (or even being redirected to https when originally fetching an http URL), it fails with an ExecutionException due to some NulPointerException.
The cause is that one has to enable https capability explicitly for the underlying Jetty HttpRequest being able to do https.
This pull request fixes this.


jonesde:
Looks good, thank you, I hadn't tried this yet over https.

PR #155, opened 2016-11-30
Improved compare method

zhangshifeng:
Fixed exception when create a new example.


jonesde:
I'd prefer to not add an additional library just to do this. What is the issue you saw or more generally the reason for this change?

zhangshifeng:
Go to https://demo.moqui.org/apps/example/Example/EditExample, click 'New Example' button, you will see an exception.

zhangshifeng:
Updated.
Apache commons-lang was removed.

jonesde:
Thank you, I see the issue that the EditExample screen uses this method with a null value which used to be supported and wasn't after the move from Groovy to Java for this method so adding back compare support for a null field is needed. Your latest changes look good but I already had some slightly simpler ones (not as good as yours in some ways, but simpler) so I committed those.

PR #157, opened 2016-12-04
Fixed ：a custom filter, init method executed twice

yanqiushi:
Fiedx：When I was in moquiDevConf.xml File to add the < filter >, the filter init method executed twice:

the first time：In MoquiContextListener.groovy  filter.init(filterConfig).
the second time：When the first request this filter will executed again.


jonesde:
Thank you, I'll test this more but the change looks sound (just register the filter and don't initialize it, let the Servlet container do that later because it will anyway). I'd prefer to opposite where the filter is initialized as part of initial startup, but never found a way to do that and forgot to loop back around and change the code to something more like this. Thank you again for bringing it up and providing the patch.

BTW, I forgot to ask you to add yourself to the AUTHORS file (in both sections, for Copyright Waiver and Grant of Patent License) to cover IP clearance for your contribution. Would you add yourself following the pattern there and submit another pull request?

yanqiushi:
Sorry ,my English is not good. Servlet container will be initialized it ,when it first uses this filter. I am very happy to add myself to AUTHORS file . thank you!

jonesde:
Thank you for the AUTHORS pull request.

PR #161, opened 2016-12-05
Fix the problem that cannot update form response

zhangwei1979:


jonesde:
Thank you, looks good. This was never updated after the switch to MNode.

Issue #183, opened 2017-02-05
Support sort on text field on Data Document


shendepu:
The field type in Entity definition is mapped to different elastic search field type. `id`, `id-long` and `text-indicator` is mapped to `keyword`, other text type like `text-medium` is mapped to elastic search `text`. But by default elasticsearch text type is not sortable. 

Is it possible to added a field `keyword` (or `esTypeMapping` where may contain `keyword,range` of elasticsearch field type separated by comma) in entity `DataDocumentField`, so component `moqui-elasticsearch` can know what additional field type need to be created when creating index mappings.

shendepu:
And if this approach is ok, I can implement it in both framework (extend DataDocumentField definition) and moqui-elasticsearch.

jonesde:
How to configure it is one question, but the first question is what is the solution in ElasticSearch to this issue. One problem with the keyword type is that it is not tokenized for search so only matches on the full string. One possible solution to this is something like this:

https://www.elastic.co/guide/en/elasticsearch/reference/current/multi-fields.html

This adds a little complexity (ie having to sort on 'city.raw' when the field is of a type with both) but makes it much more flexible. 

I've thought about this a bit over time and looked a little into options in the latest ElasticSearch, unfortunately the mapping types have changed over time and this problem introduced with more recent versions of ES.

shendepu:
Yes, multi-fields in elasticsearch is the way to go. text type field in entity definition still has the `text` type which is for full text search. 

```
 {
  "city": {
    "type": "text",
    "fields": {
      "keyword": { 
        "type":  "keyword"
      },
      "range": {
        "type": "integer_range"
      }, 
      ...
    }
  }
}
```

but key in fields is more like `keyword`, `range` which can tell the type it refers to. https://www.elastic.co/guide/en/elasticsearch/reference/5.2/fielddata.html It seems not handy to use `city.keyword` to do sorting.

shendepu:
Besides of additional field data types, there are also common use cases:

- Use `html_strip` character filter to strip out HTML elements if field text is html
- Analyze field text in different language other than English.

these two use cases are essentially how to define custom analyzer on Field, the analyzer need to exist first. It will need to add custom settings to customize analyzer.

[...]

PR #185, opened 2017-02-13
Added DataDocumentField.@esDataTypes to support adding additional fie…

shendepu:
…ld datatypes in mapping, see https://github.com/moqui/moqui-framework/issues/183


shendepu:
I also created [PR](https://github.com/moqui/moqui-elasticsearch/pull/6) to utilize this field. 

I did local test, it works.

shendepu:
see #187 which is more extensible than this one.

PR #186, opened 2017-02-14
Data document field mappings

shendepu:
The new entities are for https://github.com/moqui/moqui-elasticsearch/pull/7

There is one problem on DataDocumentField seed data, a dataDocumentFieldId shall be pre-defined in order not to create again and again when load data. 

And currently clone#DataDocument does not copy DataDocumentFieldMapping.


shendepu:
Adding a dataDocumentFieldId in DataDocumentField as the only one primary key is a breaking change for current data document seed data. 

I don't know how big impact on existing production applications. mantle-udm and mantle-usl is easy to deal with. 

Alternative is to use dataDocumentId and fieldPath as primary keys in entity DataDocumentFieldMapping, while it looks verbose.

shendepu:
Created a new PR #187 which does not change DataDocumentField entity, so does not break anything.

PR #193, opened 2017-02-27
Added sortable field on entity DataDocumentField

shendepu:


shendepu:
As we discussed, adding a `sortable` on DataDocumentField which is more generic. The corresponding moqui-elasticsearch implementation is in PR https://github.com/moqui/moqui-elasticsearch/pull/8

PR #195, opened 2017-03-10
Move message to pos. where it is not displayed when we create new account

oandreyev:
Generally I don't expect message "Password updated for user ..." when create new account in front-end. I would prefer to move this message into position where it appears only if we try to change existing password.


jonesde:
Thank you Oleg, looks good

PR #200, opened 2017-03-23
Make ScreenUrlInfo.isPermitted pass current stack to aefi.isPermitted.

chunlinyao:
ScreenUrlInfo.isPermitted return true when parent screen has a deny but sub screens has allow. This PR fixed the issue.


jonesde:
Looks good, thank you Yao. I was going to do something very similar, even going to the extent of making the isPermitted method status to force passing in everything it needs (though didn't try that to see how many new parameters there would be).

PR #201, opened 2017-03-24
Set email to utf-8 before set content, so non-english characters in b…

shendepu:
…ody html dispalys well


jonesde:
Thanks Jimmy, looks good.

PR #209, opened 2017-04-25
Use localized service to define regions in drop-down, fall back to de…

jenshp:
A way to add localized services, as a way to have localizations be able to redefine certain behaviours through configurations. This is useful for example in the case of defining how to determine the geoId options to choose from based on the country, and have the result be country-specific.
An example that uses this functionality is visible at https://github.com/Moitcl/moqui-chile

### DIFF
--- a/framework/entity/BasicEntities.xml
+++ b/framework/entity/BasicEntities.xml
@@ -149,6 +149,7 @@ along with this software (see the LICENSE.md file). If not, see
             <moqui.basic.Enumeration description="Geo Group Member" enumId="GAT_GROUP_MEMBER" enumTypeId="GeoAssocType"/>
             <moqui.basic.Enumeration description="Region of a Larger Geo" enumId="GAT_REGIONS" enumTypeId="GeoAssocType"/>
             <moqui.basic.Enumeration description="Administrative City" enumId="GAT_COUNTY_SEAT" enumTypeId="GeoAssocType"/>
+            <moqui.basic.Enumeration description="Division of Region for PostalAddress Purposes" enumId="GAT_POSTALADDR_DIV" enumTypeId="GeoAssocType"/>
         </seed-data>
     </entity>
     <view-entity entity-name="GeoAssocAndToDetail" package="moqui.basic">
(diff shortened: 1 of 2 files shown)


jonesde:
What is it that you're trying to do? This seems overly complex and configuring services to run by locale rather than making services more locale aware (or the data they use more locale aware) by configuration or other is generally preferable. I don't know what sorts of services you are plugging in with this mechanism, but perhaps it could work in a simpler way, and one that doesn't require an additional entity in the framework that isn't used for anything else. I don't have time now to review the moqui-chile code, so apologies for the quick response, but this seems to be something that would be handled by  GeoAssoc records based on the country (or even a second level based on an in between Geo based on the country).

jenshp:
What could work fine is to add a new GeoAssoc that is specifically aimed at indicating what Geo records should be used in PostalAddresses for a given country. If none of these associations is found, the code would fall back to look at GAT_REGIONS associations.
I will now submit changes implementing this.

jonesde:
Why not just use GAT_REGIONS? Right now those are only used for drop-downs anyway and if I understand correctly from what you described about addresses in Chile provinces aren't used but municipalities are. If that's the case I don't see a need to get hung up on terminology, just have a list of municipalities instead of states or provinces in the GAT_REGIONS associations for Chile.

jenshp:
OK, that should work. The reason to proposing it that way is that in some circumstances I need to access the geographical groupings in between country and municipality. I can work with defining a different GeoAssocType for these other geographical regions and that way we do not need to change the framework code.

jonesde:
Again, why not use GAT_REGIONS? If you have both provinces and municipalities associated with a country then you can always filter by the geo type and not just on the geo assoc type.

jenshp:
As we will have (for Chile) one relation bewteen country and each Region, as well as between country and Municipality, we would need to specify the geoTypeEnumId in the transition that calls the get#GeoRegionsForDropDown so it will only return that type and not mix up Regions with Municipalities. The geoTypeEnumId should be GEOT_STATE, so it works in the base cases (USA and Canada).

If there is no practical way to define the geoTypeEnumId dynamically based on the country, we would need to define the Municipalities twice for Chile: one time as type Municipality which is the official geographical subdivision and would be used to filter in other cases, and one time as type GEOT_STATE as for postal addresses to work.

jonesde:
It seems like you have two choices:

1. keep it simple and get over terminology so that you get the structure and functionality you want with no changes; in other words, just put the municipalities in as states if that is the main thing used and what should show in the drop-down
2. worry about terminology and make a bunch of changes just so a few letters look nicer in the code

Do you really need anything other than municipalities in Geo records, ie are you doing anything with the Chile province records or do you just need municipality records? You can also use a different geo type, or even geo assoc type, for regions other than municipalities if you really want them in there for some reason.

If you want more than one drop-down that is another issue, but from what you've written that doesn't seem to be the case.

jenshp:
Indeed. I will go with the simple way for the postal addresses, and add a different geo assoc type for the other Geo records, which are needed for other purposes.

PR #225, opened 2017-07-03
allow -Pconf= to override configuration file while using gradle

dhcdata:
In [Running and Deployment Instructions](http://www.moqui.org/framework/docs/RunDeploy.html#CommonAlternateSpecifyaConfigurationFileonCommandLine) a  _Common Alternate_ to provide a configuration file at the command line is provided for `java -jar moqui.war conf=xyz`

this PR allows for the same for the `gradle task`
`./gradlew runProduction -Pconf=xyz`

I don't know that this is the best practice, but since the option is there when running the war file directly, it might as well be there when running the `gradle task`

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -1,12 +1,12 @@
 /*
  * This software is in the public domain under CC0 1.0 Universal plus a
  * Grant of Patent License.
- * 
+ *
  * To the extent possible under law, the author(s) have dedicated all
  * copyright and related and neighboring rights to this software to the
  * public domain worldwide. This software is distributed without any
  * warranty.
- * 
+ *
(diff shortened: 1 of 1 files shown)


jonesde:
It is more likely we'll go in the direction of removing the run and runProduction tasks altogether. I stopped using them once the Gradle Daemon came around, and is now on by default. If you use 'gradle run' and then ctrl-c to stop it kills the daemon. It isn't that had in dev to type 'java -jar moqui.war' or for production to do the same plus a conf argument.

I would not at all recommend running through Gradle in production, it adds a few hundred MBs of memory overhead.

PR #226, opened 2017-07-06
add WEB-INF/classes so init properties are picked up

dhcdata:
Without looking further into it, `StartClassLoader.getResource("resourceName")` needs `resourceName` to be prefixed with `WEB-INF/classes` in order to pull the properties from `${moquiBase}/MoquiInit.properties`


jonesde:
That isn't how the classpath works, WEB-INF/classes is a directory that would go on the classpath but WEB-INF is never a package or path in the classpath.

What is the issue you were seeing? If something isn't working please create an issue and describe the behavior you are seeing and how it is different from what you would expect.

PR #234, opened 2017-07-15
Add is-aggregate attribute to specify if a function is an aggregate function

chunlinyao:
When not specified determined by if function is in the aggregate functions list.
Allow use other aggregate functions supported by database.


jonesde:
Looks good, thank you.

Issue #49, opened 2016-04-22
Conversion failure with auto-parameters in other locales


jenshp:
I am seeing following problem while using the spanish locale (relevant issue is that decimal separator is ',' and thousands separator is '.':
When invoking the service TaskServices.update#Task in mantle-usl, the auto-parameters for the in-parameters gets the decimal numbers wrong. These decimal numbers come directly from the Databse, so are not in the user's locale format. When checking the parameters (`checkConvertType` method of `org.moqui.impl.service.ServiceDefinition`), the line:
`BigDecimal bdVal = eci.l10n.parseNumber((String) parameterValue, format)`
assumes that the parameter is in the locale's format. But when the value comes directly from the DB because it is filled in because of the auto-parameters tag, a value like 1.5 is converted to 15 (eliminating what is considered a thousands separator according to the locale).

The solution would be to either:
1. format the parameters coming from the DB into the current user's locale, or
2. identifying when a parameter comes from the DB (it would have the entity-name and field-name attributes) and in that case use a standard parsing instead of going through the L10nFacadeImplementation.

I'm not sure which way would be more suitable.

jenshp:
A related issue is that apparently the number validation used does not take the locale into account. So, in HiveMind when I try to incorporate an hour formatted according to the curent locale (spanish) in hm/Task/EditTimeEntries, it gets rejected by the jquery validate plugin.
If I want to add an entry with 1,2 hours, I have to use 1.2 to get through the Javascript checking but then it would be recognized as 12 hours instead of 1.2 (due to the locale configuration).

jonesde:
I apologize for now response on this yet, it is indeed an issue. The first priority would be a working round trip to/from the database. The JavaScript checking is another issue, there may be a locale option for jquery-validation library that needs to be set.

jonesde:
Closing as part of general cleanup for the move to HiveMind and for no recent activity on this issue. If this comes up again or anyone wants to pursue it feel free to create a request on moqui.org. For more information see:

https://www.moqui.org/m/docs/moqui/Community+Guide

PR #235, opened 2017-07-17
Format BigDecimal values in hidden fields and drop-down options according to locale

jenshp:
When received and processed in a transition, these field's values will be treated as if in user's locale, so this change is required for a DB roundtrip through the browser to work smoothly. This solves the server part of https://github.com/moqui/moqui-framework/issues/49
It affects the value of hidden fields and options, as well as the current value in drop-down fields. The developer should, however, be locale aware if using the expandable key and text attributes of tags like entity-option and list-option, adding "${ec.l10n.format(keyName, null)} instead of just "${keyName}" for the code to work using a locale with comma as decimal separator. Making this aspect transparent to the developer would require to modify the behavior of ResourceFacade.expand(...) to make the transformation into the locale-aware format, which could affect other uses of that method.

### DIFF
--- a/framework/src/main/groovy/org/moqui/impl/screen/ScreenForm.groovy
+++ b/framework/src/main/groovy/org/moqui/impl/screen/ScreenForm.groovy
@@ -1001,6 +1001,8 @@ class ScreenForm {
                             if (keyAttr != null || textAttr != null) {
                                 addFieldOption(options, fieldNode, childNode, [entry:listOption], ec)
                             } else {
+                                if (listOption instanceof BigDecimal)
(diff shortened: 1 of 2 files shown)


jonesde:
Honestly, I'm never going to merge this. It is a partial solution to a much bigger issue that needs a lot more effort for testing different scenarios (which need to be defined) and overall handling of numbers with comma as the decimal separator. Even supporting this for only BigDecimal is inadequate, what about Double and other decimal number objects? What about pretty much every other instance of call to the toPlainString() method? There are lots of them about.

PR #187, opened 2017-02-20
Added entity DataDocumentFieldMapping and DataDocumentIndexSetting, t…

shendepu:
…o support https://github.com/moqui/moqui-elasticsearch/pull/7

PR #237, opened 2017-07-18
Move Jackrabbit initilization code to preFacadeInit method since it i…

shendepu:
…s required to be up before ResourceFacade initilization. And also return until the port is up.


shendepu:
@jonesde can you take look at if this PR is ok?

jonesde:
Frankly the JackRabbit integration is a very low priority, currently #187 on my list of priorities (that's a joke, but seriously I have a heck of a lot of other stuff to worry about and nearly no help doing it). I merged this after a quick look, not a thorough review, basically as long as it doesn't break anything else I'm fine with it.

PR #241, opened 2017-07-31
Fixed location has an extra '/'; Update move and delete method that u…

shendepu:
…ses session with session.save() to save the changes

### DIFF
--- a/framework/src/main/groovy/org/moqui/impl/context/reference/ContentResourceReference.groovy
+++ b/framework/src/main/groovy/org/moqui/impl/context/reference/ContentResourceReference.groovy
@@ -57,7 +57,7 @@ class ContentResourceReference extends BaseResourceReference {
 
         this.repositoryName = repositoryName
         this.nodePath = node.path
-        this.location = "${locationPrefix}${repositoryName}/${nodePath}"
+        this.location = "${locationPrefix}${repositoryName}${nodePath}"
         this.theNode = node
         return this
     }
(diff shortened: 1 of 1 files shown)


shendepu:
@jonesde I know jackrabbit support is low priority. This fix is pretty straight forward
1. `nodePath` starts with `/`, the `location` composition does not need the `/` after `repositoryName`
2. In the `move` method, the the first empty string of `nodePathList` shall be striped out before calling `findDirectoryNode`
3. In the `move` method, it shall call `session.save()` to persist the change 
4. In the `delete` method, the `curNode.remove()` does not work. I guess it might have something to do with session, so use the session to remove the node instead. 

I have done the test locally with jackrabbit server.

jonesde:
The changes in this for items 1 and 2 on your list change the convention to require that a node path starts with a '/' instead of what was supported in the past where the node path should not start with a forward slash ('/'). This is a non-backward compatible change and would break configuration for those already using it and have a configuration with no '/' at the beginning of the node path.

A '/' at the beginning of the node path could be supported but optional with a bit of additional logic, but I see no benefit in requiring it and changing how JCR repositories must be configured. Is there one you found?

shendepu:
For the item 1, it is initialized by jcr node with `this.nodePath = node.path` where `node.path` return the absolute path that starts with '/'. It does not affect another initialization by content location 

For the item 2, actually it is same logic with `putObject` method at line `153`. The first item is empty only when it is absolute path, while it does not affect the relative path.

jonesde:
I see, thank you, I forgot the path comes from the JCR API as opposed to a configuration file. Merging.

PR #250, opened 2017-08-23
fix exception

yanqiushi:
fix NullPointerException,when openStream method return null.


jonesde:
Thank you for bringing this up. More should be done there than just check for null before closing, ie to follow the pattern of sending a not found response if the resource referenced isn't found. I made changes to do this in commit #3d5126d1.

PR #252, opened 2017-09-05
Changed field "original" of entity moqui.basic.LocalizedMessage to text-long

jenshp:
Several strings needed to translate exceed 255 characters, in particular the validation messages in mantle.order.OrderInfoServices.check#OrderPreApprove.


jonesde:
Don't like this one... if your message is longer than 255 chars better to use an artificial key and make sure they is a value in the database. You can use the 'default' locale like all the *NameTemplate localized labels.

Stepping up to text-long is 4000 chars, and for a PK field this is messy. There are 2 issues: a string so long is likely to change and invalidate translations, slows things down to have such huge keys.

Are there strings in the framework or in mantle/SimpleScreens/etc where you ran into this? I wouldn't be too surprised, not something I've tested for or reviewed much over time...

PR #253, opened 2017-09-08
Fix wrongly assigning httpPort instead of httpHost to httpsHost

jenshp:
Fix wrongly assigning httpPort instead of httpHost to httpsHost when https-host is null.


jonesde:
Thank you Jens.

PR #257, opened 2017-09-19
Adding Time Frequency Uom records to support WorkEffort scheduling

ccarlow:


jonesde:
I added a note on the related pull request (#54) in mantle-udm. I don't like the approach in general, but even at this level these records model point in time patterns and not time/frequency durations. If we did have records for something like this they should have a different uom type. That's a big if because I don't like the idea behind these records or how they are used the mantle-udm pull request, more details in a comment there about that.

ccarlow:
I almost created a separate UOM type for them but didn't want to introduce too many changes until reviewed.

Initially I just wanted to schedule a daily task which was easy enough to do using some of the mantle-usl changes since it only had to reference a timePeriodTypeId that had a periodLength set.

Things got more complicated when I realized that holidays like Columbus day couldn't be handled as easily by scheduling annually which led me to examining the TemporalExpression capability in Ofbiz and attempting to replicate it in Moqui.

I'm open to any changes or abandoning the concept if there is a better way.

jonesde:
Based on other discussion for now we'll avoid these, so closing this PR.

Issue #259, opened 2017-10-02
subscreens-item parameter


ccarlow:
Should subscreens-item support "parameter" child nodes?

This potential need was encountered while implementing ProductionRun screens which were based on HiveMind's Project files.  

The HiveMind Project/EditUsers.xml file seems potentially reusable by ProductionRuns.  It could possibly be moved to a WorkEffort directory under SimpleScreens and included as subscreens-item by both ProductionRun and Project.  The only thing that need would need to be supported is passing the roleGroupEnumId as a parameter with a default value to differentiate the role list between Project and ProductionRun.

Does this seem reasonable or would it be better practice to just create copies of files?

jonesde:
From my experience over the years with reuse there are cases where it is very useful to avoid redundancy, and cases where you end up parameterizing too much or doing things like extending a form that extends another form, and then the complexity causes more problems than the extra redundancy.

If I understand correctly what you're talking about here the best approach is to have a separate screen that reuses the original in a more granular way such as including sections or extending forms from the other screen. There are examples of both of these in SimpleScreens and PopCommerce. This pattern allows the derived screen to set parameters or even query different data in its actions, depending on how the artifacts reused from the original screen are written of course. In some cases I've made changes to forms and screen sections to support parameters for specific types of reuse.

That said, the idea of parameters on subscreens-item is interesting for potentially other uses and I'm considering merging that anyway.

PR #260, opened 2017-10-03
subscreens-item parameter child node support

ccarlow:
This pull request contains changes to support #259.  It would, for example, allow HiveMind's Project/EditUsers.xml to be moved to SimpleScreens WorkEffort/EditUsers.xml to be reused for other WorkEffort screens such as ProductionRuns and allow the roleGroupEnumId to be specified as a parameter for differentiating the role list between Projects and ProductionRuns.

Unlike the previous implementation, the roleGroupEnumId will be exposed to users as a GET parameter so the value could easily be changed to an incompatible value (such as a ProductionRun role group instead of Project for HiveMind Project screen).  If this is a concern then it the value would have to be validated before being stored.  This along with screens like this rarely ever being reusable are the only two reasons I can come up with for why this capability may be unnecessary.

If this change is deemed unnecessary then I'll just change the roleGroupEnumId value in a copy of EditUsers.xml for Production Runs.

### DIFF
--- a/AUTHORS
+++ b/AUTHORS
@@ -51,6 +51,7 @@ Written in 2016 by Scott Gray - lektran
 Written in 2016 by Mark Haney - mphaney
 Writter in 2016 by Qiushi Yan - yanqiushi
 Written in 2017 by Oleg Andrieiev - oandreyev
+Written in 2017 by Christian Carlow - ccarlow
 
 ===========================================================================
 
--- a/AUTHORS
+++ b/AUTHORS
@@ -88,3 +88,4 @@ Wrttien in 2016 by Scott Gray - lektran
 Written in 2016 by Mark Haney - mphaney
 Writter in 2016 by Qiushi Yan - yanqiushi
 Written in 2017 by Oleg Andrieiev - oandreyev
+Written in 2017 by Christian Carlow - ccarlow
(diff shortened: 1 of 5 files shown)


jonesde:
This patch looks pretty good and I'd like to merge it (regardless of my comments on a potentially better approach for the reuse pattern in #259).

Because this is you first direct code contribution to moqui-framework please add yourself to the AUTHORS file, under both the Copyright Waiver and Grant of Patent License sections. Signing through your GitHub account is the IP clearance approach we are using and is similar to the iCLA at the ASF (just a lot easier, more public, and auditable by git commit). You can do this commit in the fork you created this pull request from and then update the pull request.

There is also a minor conflict in the xml-screen-2.1.xsd file that I was going to resolve manually, if you do a git pull from upstream/master before signing you'll see this.

jonesde:
After merging this I noticed some issues, not sure what happened or how this would have even compiled in your fork. 

In ScreenDefinition.groovy line 1030 it tries to set 'hasRequired' which does not exist in the SubscreensItem class. Commenting that for now, might be useful in the future for optimization like in the ScreenDefinition class but isn't needed now.

In ScreenRenderImpl.groovy line 1592 there was an open curly brace for the if body that was not closed. I removed that as it is not needed for the single statement if body.

jonesde:
One more issue: the subscreen parameters are only used in the getMenuData() method which is specific to the vapps/vuet mode and there is no expandParameters() method (like in ResponseItem and elsewhere) and no calls to it for the general parameter expansion in screen URLs and such.

It's a bad morning and I don't want to deal with this right now, so reverting the merge commit.

PR #263, opened 2017-10-10
EntityDataDocument.getDataDocuments() convert `data-time` field of en…

shendepu:
…tity value (java type is Timestamp) to  java type `long`, so that it should avoid time zone issue when indexed by elasticsearch. Should resove issue https://github.com/moqui/moqui-elasticsearch/issues/3.

The underlying logic is: for a field with type `data-time` with value `1507600800000` which is time `2017-10-10T10:00:00+08:00 ` and `2017-10-10T02:00:00+00:00`, when elasticsearch index value with `Timestamp`, it gets string of the timestamp which is `2017-10-10 10:00:00`, since it is string, it will be indexed as UTC time which will become `2017-10-10 10:00:00+00:00 ` (= `2017-10-10T18:00:00+08:00 `).

By converting it to long value, it will be indexed by elasticsearch with format `epoch_millis`, therefore avoid the timezone issue.

### DIFF
--- a/framework/src/main/groovy/org/moqui/impl/entity/EntityDataDocument.groovy
+++ b/framework/src/main/groovy/org/moqui/impl/entity/EntityDataDocument.groovy
@@ -249,7 +249,9 @@ class EntityDataDocument {
                 for (String pkFieldName in ddi.primaryPkFieldNames) {
                     if (!ddi.fieldAliasPathMap.containsKey(pkFieldName)) continue
                     if (pkCombinedSb.length() > 0) pkCombinedSb.append("::")
-                    pkCombinedSb.append((String) ev.getNoCheckSimple(pkFieldName))
+                    Object pkFieldValue = ev.getNoCheckSimple(pkFieldName)
+                    if (pkFieldValue instanceof Timestamp) pkFieldValue = ((Timestamp) pkFieldValue)?.getTime()
+                    pkCombinedSb.append((String) pkFieldValue)
                 }
                 String docId = pkCombinedSb.toString()
 
@@ -285,6 +287,7 @@ class EntityDataDocument {
                             for (int i = 0; i < fieldAliasList.size(); i++) {
                                 String fieldAlias = (String) fieldAliasList.get(i)
                                 Object curVal = ev.get(fieldAlias)
+                                if (curVal instanceof Timestamp) curVal = ((Timestamp) curVal)?.getTime()
                                 if (curVal != null) primaryEntityMap.put(fieldAlias, curVal)
                             }
                         }
@@ -514,6 +517,7 @@ class EntityDataDocument {
                     for (int i = 0; i < fieldAliasList.size(); i++) {
                         String fieldAlias = (String) fieldAliasList.get(i)
                         Object curVal = ev.get(fieldAlias)
+                        if (curVal instanceof Timestamp) curVal = ((Timestamp) curVal)?.getTime()
                         if (curVal != null) parentDocMap.put(fieldAlias, curVal)
                     }
                 }


jonesde:
Good idea. BTW, if instanceof is true the value is not null so no need for the '?.' operator.

jonesde:
I merged this and then considered it more as DataDocument results are not only used for ElasticSearch.

I was unable to reproduce the time zone issue in local tests, the date and time came back as expected. This was with ElasticSearch embedded. We are sending Timestamp objects to ElasticSearch so it has the object form with the millis internal representation. I haven't seen any evidence of converting the date to a string before indexing.

For the combined PK string a long value makes sense, I'll leave that in there and comment the other two until I see a need for the change. Do you have a test case to reproduce the issue? My test case was to run the mantle-usl tests then look at an order document (in the mantle_sales index/alias, actual index in ES is mantle_sales_order_item) through the document Search screen in the System app.

jonesde:
BTW, the timestamps come back fine in my local tests through the ES rest interface, ie:

curl http://localhost:9200/mantle_sales_order_item/_search?pretty

shendepu:
I also used ElasticSearch embedded, and server timezone is "Asia/Shanghai". 

it appears with fixed number of hours difference. it was my guess that elasticsearch use string to do index.

I did a quick test again by indexing document with and without converting. timestamp `1507618515481` represents "2017-10-10 14:55:15.481 +08:00", below result is from `viewDataDocument` 

Not converting `Timestamp` to `long`
```
"entryDate": "2017-10-09 23:55:15.481",
```

Converting `Timestamp` to `long`
```
"entryDate": 1507618515481
```

It seems elasticsearch use the string form of value for `entryDate` internally when not converting, and search on the result returns unexpected data.

I at early thought about that is more ealsticsearch related, and tried to fixed it when index#DataDocument which recursively traverse `documentList` and check every field whether it is `Timestamp`. But that is not performant and there might be many custom data feed recieve services and all of them would require to do the convertion.

shendepu:
In terms of test case, I had a specific chat application using elasticsearch search after with only one sort field `entryDate`, actually it is using `CommunicationEvent` entity. After this commit, I get expected result scrolling backward and forward of the chat history.

Have you checked the value of the timestamp if there is difference? elasticsearch seems store date in UTC time internally. so to reproduce it, the server needs to be in a different timezone. 

Which mantle_usl test case you have looked? I may double check it myself.

shendepu:
In my application with search after, when not converting, the endCursor is the last hit.sortValues() which only contains `entryDate`, and they are different with `7` hours.

image

In above https://github.com/moqui/moqui-framework/pull/263#issuecomment-335385827, 
```
"entryDate": "2017-10-09 23:55:15.481"
``` 
does not change even if I changed timezone of logged user, so underlying value seems not `Timestamp`. If simple compare to `2010-10-10 14:55:15.481 +08:00`, it is `15` hours, but time difference to UTC is `8` hours. then remaining  is `7` hours which is same with above. 

I don't understand why there is `7` hours there and where it comes from. even for search after in above pic, the endCursor is the `sortValues` of last hit and there should be same with the value of `entryDate` of last item.

jonesde:
When you say the 'server timezone' is Asia/Shanghai do you mean that is the Java default TimeZone? This is now set in Moqui on startup using the 'default_time_zone' property specified in a Moqui Conf XML file or from a default_time_zone environment variable (for running in Docker, etc). The Java default TimeZone is displayed on the Dashboard screen of the System app (/vapps/system/dashboard) in the System Information section at the top of the right column.

If you look at a document from ElasticSearch in the System app or through the ES REST interface (like the curl command I included) what you see is a JSON representation of the objects (in the System app that still uses the Groovy JSON API) and either way the text representation is generated based on the Java default TimeZone, not the current user's time zone (it doesn't go through the L10nFacade like normal date/time display and parsing in the applications).

If you are sending data to Moqui using a REST call, and getting data back with a REST call, you could have issues there with time zones. That adds an extra layer on top of what happens with ElasticSearch on the server side so is a more complex test scenario and you'd have to validate the Timestamp values at each step.

The simpler case I used for testing this was to just run 'gradle loadsave runtime/component/mantle-usl:test' then run Moqui with something like 'java -jar moqui.war' and in the document Search screen in the System app search for * in the 'mantle_sales' index and view one of the order documents, looking specifically at the entryDate or placedDate fields.

shendepu:
I understand how the `15` hours difference come. `MoquiDefaultConf.xml` defines `default_time_zone=US/Pacific` which is `UTC -7`, while `Asia/Shanghai` is `UTC +8`  

After run gradle loadsave runtime/component/mantle-usl:test, below is the detail of `Sales Invoice [MantleSalesInvoice] 100002` in `mantle-sales` index

```
{
    "_entity": "invoices",
    "invoiceId": "100002",
    "invoiceTypeEnumId": "InvoiceSales",
    "invoiceTotal": "44.88",
    "appliedPaymentsTotal": "44.88",
    "unpaidTotal": "0",
    "currencyUomId": "USD",
    "fromPartyId": "ORG_ZIZI_RETAIL",
    "toPartyId": "CustJqp",
    "invoiceDate": "2017-10-10 17:41:23.979",
    "dueDate": "2017-11-30 17:41:23.979",
    "statusId": "InvoicePmtRecvd",
    ... 
}
```
The `invoiceDate` is `2017-10-10 17:41:23.979` (supposed to `US/Pacific`, timestamp `1507682483979`) and it does not change on logged user timezone setting. And I also logged the class name of `invoiceDate` for `docGr.getSourceAsMap()` of the search hit for invoice `100002`
```
unpaidTotal:  0, java.lang.String
fromPartyId:  ORG_ZIZI_RETAIL, java.lang.String
dueDate:  2017-11-30 17:41:23.979, java.lang.String
invoiceDate:  2017-10-10 17:41:23.979, java.lang.String
invoiceTotal:  44.88, java.lang.String
_entity:  invoices, java.lang.String
currencyUomId:  USD, java.lang.String
statusId:  InvoicePmtRecvd, java.lang.String
customerClassId:  CustConsumer, java.lang.String
invoiceId:  100002, java.lang.String
invoiceTypeEnumId:  InvoiceSales, java.lang.String
toPartyId:  CustJqp, java.lang.String
appliedPaymentsTotal:  44.88, java.lang.String
```
The `invoiceDate` is string not Timestamp.

Then I do before query test on `mantle-sales` index to see if the invoice `100002` in the search result, (true: show in the list, false: not in the list)
1. invoiceDate: ["2017-10-10 17:41:23.979" TO *] : true
2. invoiceDate: [* TO "2017-10-10 17:41:23.979"]: true
3. invoiceDate: [* TO 1507682483979]: true
4. invoiceDate: [1507682483979 TO *]: false

The question raise why 3 fails. timestamp `1507682483979` is `2017-10-11 00:41:23.979` in UTC time. so I try to run query with timestamp `1507657283979` (which is `1507682483979 - 3600 * 1000 * 7) which is ,
1. invoiceDate: [* TO 1507657283979]: true
2. invoiceDate: [1507657283979 TO *]: true

It appears that elasticsearch use date string `2017-10-10 17:41:23.979` and index it as UTC time `2017-10-10 17:41:23.979`, but it actually should 2017-10-1` 00:41:23.979` of UTC time. If search on date by date string, it is fine since date string is treated as in UTC time. but if search on the date with timestamp long value, then there is timezone issue. 

If index the date field with timestamp long value, then the issue is gone. I guess elasticsearch does not use `Timestamp.getTime()`, but simply use `Timestamp.toString()` when index field with value of `Timestamp`.

shendepu:
@jonesde Have you got a chance to look at this again? I think the logic of time difference issue when index with elasticsearch is clear now.

jonesde:
No, from your write up it isn't clear at all what's actually going on in ElasticSearch. Some tests I did didn't fit with what you wrote up as an explanation of ES behavior, but the symptoms are certainly there.

After hours more wasted on ElasticSearch weirdness (poorly documented weirdness in spite of their large and overall decent documentation) it seems that ES does some odd things with Java types it doesn't know about, including Timestamp and BigDecimal. It appears that it simply calls toString() on them (and then treats the String as a date, double, etc when doing operations on them such as searches and aggregations).

Converting a Timestamp to a Long (millis since epoch) does handle this as it is a type ES knows about, as does converting BigDecimal to Double. This is very specific to ElasticSearch and as I mentioned before such things don't belong at the DataDocument level which is a more general tool and not just for feeding ElasticSearch. It could be used for SOLR or other search utilities, and is used for other things like notifications where we don't want the types mucked around with.

To handle this specifically for ES there are now some changes in the moqui-elasticsearch component to do nested type conversions before sending document Maps to ES.

The results coming back from ES are much better with this change too. Instead of getting String values for Timestamp and BigDecimal fields we now get back a Long (millis since epoch) for Timestamp and an actual number in the source Map or JSON for BigDecimal fields.

There may be other types that need extra attention but in reviewing various documents coming back from ES these 2 are the main ones.

shendepu:
Thanks David. This is more elasticsearch issue, and commit in moqui-elasticsearch is in better place.

PR #266, opened 2017-10-13
Wiki Blog - process wikiPageId in Wiki blog services

oandreyev:
Required after changes in the entity.

### DIFF
--- a/framework/service/org/moqui/impl/WikiServices.xml
+++ b/framework/service/org/moqui/impl/WikiServices.xml
@@ -523,7 +523,7 @@ along with this software (see the LICENSE.md file). If not, see
     <service verb="create" noun="WikiBlog">
         <in-parameters>
             <auto-parameters entity-name="moqui.resource.wiki.WikiBlog" include="nonpk">
-                <exclude field-name="smallImageLocation"/><exclude field-name="blogLocation"/></auto-parameters>
+                <exclude field-name="smallImageLocation"/></auto-parameters>
             <parameter name="wikiSpaceId" required="true"/>
             <parameter name="title" required="true"/>
             <parameter name="summary" allow-html="safe"/>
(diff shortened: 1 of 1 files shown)


jonesde:
WikiBlog now has a wikiPageId field, tying directly to a WikiPage record. Rather than using the title as the path (or page name) when creating a WikiBlog the wikiPageId should be specified explicitly. That will simplify the create and update services, and handle the situation where the page name or path (via parent page) is changed in the wiki edit screen.

In the UI for a WikiBlog record it should have a simple drop-down, similar to the parentPath drop-down in EditWikiPage.xml, but with a key of the wikiPageId to pass to the create# and update#WikiBlog services.

oandreyev:
> WikiBlog now has a wikiPageId field, tying directly to a WikiPage record. Rather than using the title as the path (or page name) when creating a WikiBlog the wikiPageId should be specified explicitly. That will simplify the create and update services, and handle the situation where the page name or path (via parent page) is changed in the wiki edit screen.

Yes. I noticed that. Now it uses wikiPageId.

jonesde:
The reason I commented it because the changes to the services in this pull request are not adequate. 

The first line in both the create# and update#WikiBlog services calls the update#WikiPage service, and there should be no call to update#WikiPage at all in these services. To hopefully clarify, the WikiBlog.title field should only be used for the title in the blog listing, it has nothing to do with the WikiPage.pageName (and in turn nothing to do with the page path).

There should also be no blogText passed to either the create or update WikiBlog services, all of that is part of the WikiPage based on the wikiPageId passed to the create and update services.

When I mentioned the WikiPage should be selected from a drop-down I meant that explicitly, used to pass the wikiPageId to the create/update services. This would be used instead of the big text area (with CK Editor) which should be removed. Creating a blog post is a 2 step process: create the wiki page (wherever desired in the space, ie can be under other wiki pages to organize them) and then create the WikiBlog record to list it in one or more blog listings based on wiki categories.

On the edit blog page it would be good to have a link to the view wiki page, and perhaps we should add a link on the edit wiki page screen to the wiki blog record(s) referring to it. We might want a form to create both at once, but that would just be a simplified create form and I'd prefer to avoid it to avoid having another CK Editor instance to maintain.

jonesde:
An alternative to this is in now in place, commit #b17d8e58.

PR #271, opened 2017-10-26
Add mantle-braintree to list of available components

oandreyev:


jonesde:
Thanks Oleg, I forgot to do this. This patch undoes the changes for the recent releases so I just added this by hand in commit 8bc67f76.

PR #282, opened 2017-11-16
Changed "original" field in moqui.basic.LocalizedMessage to "text-long"

jenshp:
Some messages exceed 255 characters, such as approve warnings in mantle.order.OrderInfoServices.check#OrderApprove and do not fit into a text-medium field.
The "localized" fields both in LocalizedMessage and LocalizedEntityField also are of type text-long.

### DIFF
--- a/AUTHORS
+++ b/AUTHORS
@@ -82,7 +82,7 @@ Written in 2015 by Jimmy Shen - shendepu
 Written in 2015-2016 by Sam Hamilton - samhamilton
 Written in 2015 by Leonardo Carvalho - CarvalhoLeonardo
 Written in 2015 by Anton Akhiar - akhiar
-Writter in 2015-2016 by Jens Hardings - jenshp
+Written in 2015-2017 by Jens Hardings - jenshp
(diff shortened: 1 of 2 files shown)


jonesde:
I'm pretty sure I've commented on past requests to change this and that a different solution was needed. Better to open an issue than a pull request (start with the problem, not with a proposed solution).

I don't see how this would even help anything without changes to the L10nFacadeImpl.localize() method which is written for original strings that are 255 characters or less, currently it throws an exception if the original string is longer.

One thing I've considered thinking on this over time is to use just the first 255 characters as the original string, ie trim to that length before lookup instead of throwing an exception.

jonesde:
BTW, what is a test case for this, ie steps to reproduce an issue? Based on current code in L10FacadeImpl.localize() I would expect to see an exception thrown somewhere, but I'm not aware of this being the case for any of the current code in the framework, business artifacts, or applications.

If you're running into this in a custom application or with modifications to one of the Moqui Ecosystem projects (like mantle-usl, SimpleScreens, etc) that would be the place to start in terms of discussion of how to do it differently.

PR #299, opened 2018-01-07
Unit tests for sub-select

chunlinyao:
This is an simple unit tests for sub-select.
Sometime generated WHERE clause for sub query use `entity-alians.field-name`,  It should be `field-name` only.
The unit test contains two case, one will fail, another will pass.


chunlinyao:
Don't known why test can passed on my local PC, but failed on Travis.
@jonesde Do you have some idea about it.

PR #308, opened 2018-03-05
Handled case where jndi-jdbc is defined without a inline-jdbc node.

aabiabdallah:
The system is throwing a NPE when a jndi-jdbc node is defined without a inline-jdbc node.


jonesde:
Thank you Ayman, I obviously haven't tested a JNDI based configuration for a long time! In commit 8e31b715 I made this line null safe and added some better checks lower down to handle missing inline-jdbc and jndi-jdbc elements.

Could you test these updates and see if they fix your issue?

For future reference the AUTHORS file needs to be 'signed' in both sections (CC0 and patent).

aabiabdallah:
Welcome David. Commit 8e31b71 resolves adds the necessary checks.

PR #310, opened 2018-03-12
Fix bug when fieldsToSelect contains all fields but not in same order…

chunlinyao:
… as allFieldInfoArray the query result incorrect.

Reported by ZhenZhao Wei


jonesde:
It looks like this is caused by the optimization in EntityQueryBuilder.java in the makeSqlSelectFields() method around line 192 where it uses the pre-built allFieldsSqlSelect String. So yes, this logic basically matches that and this is a good fix.

PR #319, opened 2018-05-11
Pull request Moqui-util

xlreconlx:
the method simpleHttpStringRequest is overridden


danieltaylor-nz:
You need to append your name to The AUTHORS file Anderson

jonesde:
These HTTP request methods are old and I've stopped using them in newer services, basically deprecated by the RestClient class with its fluid API. Basically the idea of having a bunch of different combinations of parameters is really inefficient to build, maintain, and to learn and use when compared to an API that starts with the basics and has methods returning a self-reference to easily chain method calls.

There are good examples of this in the ShippoServices.xml, GoogleServices.xml, InstanceServices.xml, etc files.

For future reference what Daniel mentioned is correct, all contributors need to self-sign the AUTHORS file (in both of the sections, ie you'll add two lines with your name, etc) before contributions can be merged into the main moqui repositories.

PR #325, opened 2018-07-08
Added DbFormAndUserGroup view entity for User Group section on EditSurvey screen in PopCommerce Survey functionality.

nirendra10695:
Added DbFormAndUserGroup view entity for User Group section on EditSurvey screen in PopCommerce Survey functionality.


jonesde:
Some quick feedback:

- use fully qualified entity names (with package prepended)
- alias-all elements should go before alias elements; you should get a warning or error from the XSD if you have it setup properly in your editor (technically the framework doesn't care, but it keeps things more consistent and easier to spot overrides and such)
- not required, but I usually try to use 3 letter aliases (longer if needed for view entities with lots of member entities)

nirendra10695:
I changed  DbFormAndUserGroup view entity entity-alias and also  added DbFormField update/delete services in ScreenServices.

PR #328, opened 2018-08-06
Added make condition methods with varargs.

aabiabdallah:
I added a couple of methods I used while writing complex find conditions, they offer a convenient way to add multiple conditions directly without having to add them to a list first.


aabiabdallah:
Any feedback on this PR?

jonesde:
Low priority. Not needed in groovy (use the [] list syntax) and in Java can use Arrays.toList() with the {} array syntax. When I do eventually loop back around there are other things I'd like to do with a real fluent API for condition building, though that is a lower priority for me.

As for timing and feedback, not much I can do about that, my time and resources are very limited and currently overextended.

aabiabdallah:
Hello! The Arrays.toList() is actually being used for the implementation. The varargs alternative just provides a cleaner approach. As for the timing, it is totally understandable, let me know if there is anything I can go to help.

aabiabdallah:
I have updated this PR to address 2 points:

1. Set RestApi entity find limit to 100 only if pageNoLimit is not set true
2. Ensured orderBy field names are unique as this is not supported in MSSQL (see below)

> Server: Msg 169, Level 15, State 1, Line 1
> A column has been specified more than once in the order by list.  Columns in the order by list must be unique.

If you're not comfortable with the 2 new methods I've added to the EntityConditionFactoryImpl, please let me know so I can remove them.

PR #332, opened 2018-09-12
Added JAXB lib for Java 9/10 compatibility.

okroener:
XML library JAXB is missing in JDK 9 and onwards. Therefore I have added the lib to the build script in order to enable normal functionality.


jonesde:
Thank you, got another question about this in Java 10 today. I added a small variation on this in commit 1c916f1, just using version 2.3.0 to avoid the beta version, and the Travis tests failed with that version.

okroener:
You are welcome! Good idea to change to v2.3.0. Thanks.

PR #336, opened 2018-10-12
E-Mail issue with some certificates including gmail, etc.

okroener:
Hi David,

we had some problems with the e-mail service when we used our own server or tried to use gmail as a smtp provider.

I have made a quick fix, that allows every domain to be trusted. My idea is to extend the MailServer configuration by a checkbox like "trust hostname", what do you think?

Thanks,

Oliver


jonesde:
There are some issues with this. First is that the current code is being used in production successfully for the following scenarios:

- internal SMTP server
- Amazon SES
- GMail account (with less secure access turned on so username/password authc can be used)

Any changes to this code requires testing on all 3, and you mention having trouble with gmail but I have both test and production configurations working fine with that.

Always adding the 'mail.smtp.ssl.trust' property doesn't seem like a good idea for most use cases (too easy to a man in the middle attack). It could be something configurable with a text-indicator field on EmailServer.

Calling the setStartTLSRequired() is something that would have to be tested for at least the 3 main scenarios mentioned above, and IIRC I removed that because it caused issues.

For these reasons I'm rejecting this pull request, though with some the configuration option the mail.smtp.ssl.trust changes might be useful, especially for those with less secure internal SMTP server (though that's not something I'd recommend).

PR #337, opened 2018-10-12
Add new gradle task 'createComponent'

telenieko:
This task allows to easily create new components like the createPlugin
task from OFBiz.

Usage:
  ./gradlew createComponent -Pcomponent=mycomponent

You can also supply your own template (must exist in myaddons.xml or addons.xml):
  ./gradlew createComponent -Pcomponent=mycomponent -PbaseComponent=moqui-component-template

NOTE: This commit adds a component to addons.xml which is a base template based on the moqui example.
If this goes into Moqui, that template or anotherone shall be put in its place.

### DIFF
--- a/addons.xml
+++ b/addons.xml
@@ -25,6 +25,7 @@
 
     <!-- Example Component -->
     <component name="example" group="moqui" version="2.1.0" branch="master"/>
+    <component name="moqui-component-template" group="telenieko" branch="master" />
 
     <!-- Moqui Tool Components -->
     <component name="moqui-atomikos" group="moqui" version="1.0.0" branch="master"/>
--- a/build.gradle
+++ b/build.gradle
@@ -538,13 +538,22 @@ task createComponent(description: 'create a new component based on the moqui exa
         compXmlFile.withWriter { w->
             w.write( XmlUtil.serialize(compXml))
         }
-
         // cleanup the example
-        file("${componentDir.path}/LICENSE.md").delete()
-        file("${componentDir.path}/AUTHORS").delete()
-        if(file("${componentDir.path}/.git").exists())
-            file("${componentDir.path}/.git").deleteDir()
-
+        File removeFilesFile = file("${componentDir.path}/remove-files.txt")
+        if(removeFilesFile.exists()) {
+            removeFilesFile.withReader { r ->
+                String fpath = ""
+                while ((fpath = r.readLine())!=null) {
+                    File f = file("${componentDir.path}/${fpath}")
+                    if(!f.exists())
+                        continue
+                    if(f.isFile())
+                        f.delete()
+                    if(f.isDirectory())
+                        f.deleteDir()
+                }
+            }
+        }
         logger.lifecycle("Created new component ${componentName}, dependent components checked: ${compsChecked}")
     }
 }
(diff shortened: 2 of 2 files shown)


jonesde:
A gradle task to create a skeleton component is an interesting idea, though components vary sufficiently that a really useful one is difficult to do.

I'm rejecting this patch because it relies on a third party component, and a better approach would be to take explicit parameters like the component name and generate the needed files... though what those would be is another question. The only obvious one is the component.xml file, but that is optional anyway and only really needed to declare component dependencies.

What might work is various gradle tasks for different types of components such as with or without a build file. The files in the generated component could be based on templates (preferably FTL files, and use FreeMarker in the gradle script to turn them into .xml, .gradle, etc files).

telenieko:
Hi,

> A gradle task to create a skeleton component is an interesting idea, though components vary sufficiently that a really useful one is difficult to do.

Yup, that's why the proposal accepts any addon as a template to use. Thus one can create custom template addons and Moqui can provide different sets of templates. -- ie: templates are just other addons.

> I'm rejecting this patch because it relies on a third party component, 

As noted on the pull request: "NOTE: This commit adds a component to addons.xml which is a base template based on the moqui example. If this goes into Moqui, that template or anotherone shall be put in its place."

So, there would not be a realiance on a third party component as the base template would be provided directly in the Moqui github org. The current reliance is temporary as a way of proof of concept.

> and a better approach would be to take explicit parameters like the component name and generate the needed files... though what those would be is another question. The only obvious one is the component.xml file, but that is optional anyway and only really needed to declare component dependencies.

It appears to me that 100% code generation would make creating template components considerably difficult. The proposed model clones a template and, if needed, does the appropiate modifications (currently changing component.xml).

> What might work is various gradle tasks for different types of components such as with or without a build file. The files in the generated component could be based on templates (preferably FTL files, and use FreeMarker in the gradle script to turn them into .xml, .gradle, etc files).

I did not want to mess much with gradle as that would require to anticipate every possible reasonable use case. On my head, the necessary template tasks shall be inside the template component itself (ie: a Groovy script or gradle task *inside the component* that can be executed after the cloning and cleanup happens).

That would keep moqui-runtime clean (it would just have a simple gradle task similar to the one here provided) and "template creators" would have no restriction on what could be done.

> Closed #337.

It would do good to community building to encourage debate/discussion before completely rejecting PRs. Specially given that the main reason for rejection (reliance on 3rd party component) is a misunderstanding as written above.

jonesde:
Pull requests are not a great tool for discussion, in this case the framework mailing list would be better (moqui@googlegroups.com). A pull request is a request to have particular changes added to the core repository which implies a request for the changes to be reviewed, tested, supported, and maintained. Changes won't go in unless they generally useful and ready to be merged, or near enough ready and interesting enough to be worth the time to clean up and/or fix. 

Please understand that there are enough pull requests and little enough resources to review and handle them that triage is a big deal. Your response here is exactly why I hate working on pull requests, lots of stuff with issues and people that get offended easily. I don't have time to nurse people through all of that or to give very much attention to every idea that comes along.

If you find your changes useful to you then you can keep and maintain it in your own repository, it doesn't need to be in moqui-framework even if that might seem more convenient initially.

I don't see a misunderstanding here, I see a pull request with changes that I don't think belong in the main code base, even if a variation on the concept might be useful.

PR #338, opened 2018-10-25
uncomplete relation causing registration to fail in popreststore

hansbak:
relation is missing the field name, i noticed more entities having the same problem?

### DIFF
--- a/AUTHORS
+++ b/AUTHORS
@@ -53,6 +53,7 @@ Writter in 2016 by Qiushi Yan - yanqiushi
 Written in 2017 by Oleg Andrieiev - oandreyev
 Written in 2018 by Zhang Wei - zhangwei1979
 Written in 2018 by Nirendra Singh - nirendra10695
+Written in 2018 by Hans Bakker - hansbak
 
 ===========================================================================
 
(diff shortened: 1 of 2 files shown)


hansbak:
the standard moqui installalation from git when register in the popreststore gives the following error:
 An error occurred: Error creating UserLoginKey {loginKey=dcedbe8ce6190d0b10457bdf75fe3f744a9beb1aee51e794d86284276858041d}: record specified does not exist [23506] Referential integrity constraint violation: "USERLOGINKEYUSERACCOUNT: PUBLIC.USER_LOGIN_KEY FOREIGN KEY(USER_ID) REFERENCES PUBLIC.USER_ACCOUNT(USER_ID) ('100000')"; SQL statement: INSERT INTO USER_LOGIN_KEY (LOGIN_KEY, USER_ID, FROM_DATE, THRU_DATE, LAST_UPDATED_STAMP) VALUES (?, ?, ?, ?, ?) [23506-197]

hansbak:
But yes you are right, this patch does not fix the problem.
the problem in PopRestSTore is that an UserAccount record gets created fine, but the insert in userLoginKey does not find the userid in UserAccount......
so sure you can reject this patch.

PR #340, opened 2018-11-01
Fixes to rest API limit and duplicate entity find order by field names.

aabiabdallah:


jonesde:
This looks good overall. In RestApi.groovy in that if clause there is a !containsKey() and !get() that are redundant and in common cases will result in 2 hash lookups, I'll remove the !containsKey() part of it. There is also a line that looks like it only had an indentation change, I'll change that back.

jonesde:
Follow up commit is a2cdf4eb

PR #352, opened 2019-02-15
Added GeoPointType enumeration type to seed data.

aabiabdallah:


jonesde:
What is the business context you're looking at for GeoPointType? To date I haven't really found a use for it, was more of a speculative part of the model.

BTW, please avoid reformatting of existing lines in patches. It makes it more difficult to pick out what actually changed, and the main reason for not formatting the entity XML files the way you have here is that the files get excessively large (not trying to win a lines of code/text contest, may take a little to get used to but more concise text makes the files easier to scan and read through).

aabiabdallah:
We're building a real-estate management system and we use the GeoPointType to distinguish between the main property location, entrances/exits, etc. I thought it would be useful to have this in the framework since this relation is defined there, and component would plugin their required values.

With respect to formating, I understand the point of view and will update the PR with the old format if its going to be accepted.

aabiabdallah:
I've restored the previous formatting and fix a minor issue with ScreenForm

jonesde:
What does the change to ScreenForm.groovy fix? It does change the behavior from using the last entity relationship found to using the first relationship found. I don't know that it will break anything, but much of the framework is this way to allow overrides by name.

aabiabdallah:
I added the break statement based on the comment in that code block, the comment says:

// to see if this should be a drop-down with data from another entity,
// find **_first_** relationship that has this field as the only key map and is not a many relationship

jonesde:
On the ScreenForm break: I researched the git history a bit to see what has been done with this over time and why. The initial commit (319d6609702a72e809f7fe5921d9a59dfc704420) has both the comment and no break statement, and going back before than (Jan 2016) would require researching the old moqui/moqui repository (before the repo split to moqui-framework and moqui-runtime), and I'm not sure I care that much. This is only used for auto form fields based on an entity definition which is mostly used in the Auto Screens of the Tools app (unless someone uses it in an application form, which can easily be done but is not generally recommended, real world screen/form designs aren't usually so simple and while automatic CrUD screens can be useful they are more technical and less usable).

Anyway, those are my thoughts on the topic. I don't think this matters much so if there are backward compatibility issues where this breaks something someone is doing it should be minor.

jonesde:
Back to the original GeoPointType topic: from a business and modeling perspective I don't know if I'd recommend using the geoTypeEnumId for something like entrance vs main location (street address equivalent), etc. This git change is fine, but that use of the field doesn't seem to align with the main principle of modeling: conceptual cardinality. The concept you are talking about does not have a one-to-one cardinality with GeoPoint so should be in a field on that entity. It is more like a concept that relates to both GeoPoint and Facility so should be on a join entity between the two.

Right now there is no join entity between Facility and GeoPoint, ie an entity named 'FacilityGeoPoint' (GeoPoint being the base or more generic entity, Facility being the higher level one so the entity would go along with the other Facility* entities). An entity like that would allow you to associate multiple GeoPoint records with a Facility record and for each you could say what type of GeoPoint it is for that particular Facility.

BTW, the current Facility.geoPointId field is meant only for the 'main' location of the Facility, ie equivalent to the street address of the facility (for a building Facility anyway).

None of that is really relevant to this particular change in BasicEntities.xml, but is an important principle in modeling and might be helpful for whatever you are implementing.

aabiabdallah:
I checked the ScreenForm older code and found that return statement wasn't there in the initial commit so this change shouldn't break anything.

With regards the GeoPoint. We're indeed following a one-to-one relation when referencing the GeoPoint. My initial comment may not have explained thoroughly enough. Thank you very much for the detailed explanation you provided anyway.

jonesde:
Now being there from the beginning actually worries me more in that it is a non-backward compatible change in behavior that could potentially break existing forms, and those forms could go back a long time. Still, probably doesn't really matter. If I were to make a design decision it would be do NOT break, find the last so that extend-entity can be used to override relationships for auto entity based form fields just like it is in other parts of the framework. The fix would be to the comment, not the code. But, I don't really care.

On GeoPoint you're missing the point of my comments... it was clear you were thinking of this concept as one-to-one with GeoPoint and my comment was that it is probably not one-to-one with GeoPoint and is not an inherent attribute of GeoPoint, is really more like an inherent attribute of a GeoPoint used for a particular Facility so is really one-to-one with GeoPoint AND Facility, ie rather than being one to one with the single key geoPointId it is more like one to one with a dual key of facilityId and geoPointId. That said, for my part and for Moqui it doesn't matter (would only matter if there was sample or seed enum data for this type, FWIW if I did put things in there they would be things inherent to GeoPoint only like geo coordinate standards and basis).

PR #356, opened 2019-04-18
Force XML and CSV sources to be interpreted using the en_US locale 

jenshp:
Force XML, CSV and REST sources to be interpreted using the en_US locale instead of the current value of the default_locale property.

### DIFF
--- a/framework/src/main/groovy/org/moqui/impl/context/L10nFacadeImpl.java
+++ b/framework/src/main/groovy/org/moqui/impl/context/L10nFacadeImpl.java
@@ -133,16 +133,19 @@ public class L10nFacadeImpl implements L10nFacade {
 
     @Override
     public Time parseTime(String input, String format) {
-        Locale curLocale = getLocale();
+        return parseTime(input, format, null);
+    }
+    public Time parseTime(String input, String format, Locale locale) {
+        if (locale == null) locale = getLocale();
         TimeZone curTz = getTimeZone();
         if (format == null || format.isEmpty()) format = "HH:mm:ss.SSS";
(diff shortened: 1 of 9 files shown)


jonesde:
Thanks Jens, overall these look like reasonable changes. Some could be simplified and avoid replacing values passed to the methods (like the Codacy review points out, I haven't started working on those cleanups but have GitHub-Codacy setup with the intent of doing this at some point). This would also reduce the patch size, ie using something like 'Locale curLocale = locale != null ? locale : getLocale();' instead of the 'if (locale == null) locale = getLocale();'.

On the overall approach I like your idea better about supporting a locale in the data file, like the standard attribute on the root element of an XML file and maybe another optional entry in the first line of a CSV file (along with entity/service name, etc). It would be nice to avoid another hard-coded value like this (in spite of all the configuration options all over the place there are still too many of these). It could default to 'en_US' if no locale specified (like other defaults elsewhere) but would be nice to support an override.

BTW, a temporary workaround that comes to mind is to set the default locale to en_US or whatever will work for the files to be loaded, then change the default locale back. For manual imports through the Tools app the current code would support it if the active user's locale is compatible with the file.

jenshp:
Yes, I noticed the checks with Codacy, so will work on that and update this pull request.

jenshp:
Just finished adding configurable options for XML, Json and CSV imports:
XML: specify locale string as entity-facade-xml.@locale or seed-data.@locale attributes
Json: specify global locale in first record together with dataType, or per-record (overrides jsonObj-wide global specification) as "_locale" value.
CSV: specify as third parameter on first line, after data-type

jenshp:
This is ready to be merged from what I can see, have been actively using it some while now.

PR #357, opened 2019-04-22
Improved the message content shown when user request for password reset

swapnilmmane:
Since, the reset password can't be used for login, added this information.


jonesde:
I don't see the point of this change, seems the next sentence in the email already says nearly the same thing: "Your current password is still valid, this reset password may only be used to change your password."

The current text could certainly be improved, perhaps with instructions to click on the Change Password link on the Login screen. If you want to work on this please start a discussion on the moqui@googlegroups.com mailing list.

swapnilmmane:
Thanks for your comment, David.

PR #359, opened 2019-05-14
Docker host conf

jenshp:
Add docker Host Conf configurations. Allows setting a wide range of parameters, including memory or CPU limits.


jonesde:
Thank you Jens, looks like a good improvement. I noticed the fix of a hard-coded value that I didn't see before, thank you for taking care of that.

PR #362, opened 2019-06-03
Add basic health check to docker image

jenshp:
Enables very basic health check, accessing /status transition.
Requires pull request https://github.com/moqui/moqui-runtime/pull/143 which enables /status for requests coming from localhost.


jenshp:
Some changes as of discussion in https://github.com/moqui/moqui-runtime/pull/143

PR #366, opened 2019-06-27
Made the taga/tagb parameters required and allowed 'HEAD' to be used

danieltaylor-nz:
users can now call:

`gradle gitDiffTagsAll -Ptaga=v1.0.0 -Ptagb=HEAD`

or

`gradle gitDiffTagsAll -Ptaga=v1.0.0`

In both cases, the diff will be between tag v1.0.0 and HEAD


jonesde:
Looks good

PR #367, opened 2019-07-03
Improved documentation for the masterName attribute in rest-api-2.1.xsd

adityasharma7:


jonesde:
This isn't quite correct, some notes:

"Name of the master definition applicable."

In this case it is the name of the master definition to use for the entity operation in a REST API. If no masterName is specified only the named entity will be used (with no master definition for related entities). I'm not sure what 'applicable' means in this context. 

"Each master definition represents specific structure of the entity."

A entity master definition is not a 'structure of the entity', it is a concept that sits on top of an entity definition to describe sets of related entities that can be used together for certain things like data export and import.

adityasharma7:
Thanks @jonesde :) Makes Sense.

> I'm not sure what 'applicable' means in this context.  

By 'applicable', I meant to say that from the list of master definitions, the name identifies the specific master definition applicable here.


> A entity master definition is not a 'structure of the entity', it is a concept that sits on top of an entity definition to describe sets of related entities that can be used together for certain things like data export and import.
Indeed. I missed this part. Each master definition represents a specific group of the data described by a set of related entities.

PR #372, opened 2019-08-02
Updated ArtifactAuthzFilterAndSet to remove warning message

aabiabdallah:
The warning was caused by a duplicate applyCond field introduced in commit bb3d234


jonesde:
Thanks for bringing this up Ayman, I made a slightly simpler change and checked on usage of the view-entity, is only for display on the artifact group and user group screens in the System app and updated those to display the authz and filter apply conditions.

PR #405, opened 2020-03-23
add elasticsearch support on windows.

chunlinyao:
Support download and start elasticsearch on windows.


chunlinyao:
Tested several days, It seems ok.

PR #411, opened 2020-04-13
Enable running embedded ElasticSearch within docker container

jenshp:
Docker runs its processes as root by default, which is not allowed by ElasticSearch. In order to enable the embedded ElasticSearch to be run, it needs to run as a different user.
This patch includes preparing the docker container by creating a user+group that owns the runtime/elasticsearch directory and upon startup checking whether the external ElasticSearch process has to be executed on behalf of the elasticsearch user, checking that:
1. the directory is owned by the elasticsearch user
2. the current user can issue a "su -c" user to execute a command as the elasticsearch user (implies that the user is root and elasticsearch user exists)

### DIFF
--- a/docker/simple/Dockerfile
+++ b/docker/simple/Dockerfile
@@ -17,6 +17,9 @@ COPY execlib execlib
 # always want the runtime directory
 COPY runtime runtime
 
+# create user for elasticsearch and chown corresponding files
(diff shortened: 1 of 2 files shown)


jonesde:
I don't know that it's the best idea to run ElasticSearch in the same container when running under Docker (some benefits to running in a separate container), but sure, might as well support it. Looks good.

PR #415, opened 2020-05-19
Added btn-type attribute for submit

zhangwei1979:


zhangwei1979:
Close it as there is no any further comments.

PR #423, opened 2020-07-07
Checked for null thruDate when logging in with a login key, removed BigDecimal method deprecated in JDK 11 

aabiabdallah:


jonesde:
Both changes look good, thanks Ayman.

PR #435, opened 2020-09-04
Fix for when "load" exceptions don't cause an error exit.

### LINKED ISSUE #434: Errors when calling load from the command line do not have an error exit value.
We are beginning to use kubernetes with moqui, and have an initContainer that calls "load" directly, before launching moqui.  During startup, the external postgresql server is not yet available(mostly because the image is still being downloaded externally), while moqui itself starts up fast, because it is all still local.  Moqui attempts to contact the postgresql server many times, prints connection refused(as it should), but then exits with 0.

I already have it fixed, there will be a pull request shortly.

eigood:
When calling "load" from the command line, and an error occurs(via some thrown Exception), make certain the virtual machine exits with a non-zero value.

This helps with kubernetes initContainer workflows.

This fixes #434.

### DIFF
--- a/framework/src/start/java/MoquiStart.java
+++ b/framework/src/start/java/MoquiStart.java
@@ -138,17 +138,20 @@ public class MoquiStart {
             initSystemProperties(moquiStartLoader, false, argMap);
             Process esProcess = argMap.containsKey("no-run-es") ? null : checkStartElasticSearch();
 
+            boolean successfullLoad = true;
             try {
                 System.out.println("Loading data with args " + argMap);
                 Class<?> c = moquiStartLoader.loadClass("org.moqui.Moqui");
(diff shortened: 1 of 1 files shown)


eigood:
That's an interesting code quality failure, about not using System.exit; the original code uses it too. :)

jonesde:
Thanks Adam, looks good to me, will merge.

Yeah, Codacy is like any other static code analysis tool, often not helpful. I set it up a long time ago with the idea of going through the results and improving code 'quality', but haven't done that seriously yet and the little I've looked at what Codacy produces I'm not seeing much worth putting effort into, and a few things like this that are just off (yeah, most is JEE code but that class is not being a bootstrap for an embedded Jetty server).

PR #440, opened 2020-09-23
Add support for try/catch/finally in XmlActions

lombardo2:
XmlActions does not have try/catch/finally tags. This is trivial to workaround using a script tag. The codebase itself does it in `Wikiservices.xml` and `SystemMessageServices.xml`. I reckon this change might have been considered before, so are there any reasons we don't have try/catch/finally tags? It should help reducing the usage of script tags and it's syntactially easier to read as if doesn't require code-switching

### DIFF
--- a/framework/template/XmlActions.groovy.ftl
+++ b/framework/template/XmlActions.groovy.ftl
@@ -305,6 +305,19 @@ ${.node}
 
 </#macro>
 
+<#macro try>  try {
+    <#recurse .node/>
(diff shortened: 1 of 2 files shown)


jonesde:
Most scripts don't need it, but no harm having it there. Just a couple of things:

1. please sign the AUTHORS file like you've done in other repos
2. the catch should also have minOccurs=0 for try/finally (yet more rare, but useful sometimes)
3. define the catch and finally elements under the try element instead of separate with a ref (have been slowly moving more toward this for better scoping of element defs, and various XML editors handle autocomplete suggestions better, have noticed even IntelliJ shows elements that aren't allowed or applicable sometimes, because they are defined globally)

PR #441, opened 2020-10-04
Added new add-unique-as database configuration attribute

aabiabdallah:
The current hardcoded behaviour was failing on certain view-entities in Oracle 12c (12.1.0.2.0)


jonesde:
Thanks Ayman, for future reference please see my follow up changes in commit ec2276e4, mainly the NPE issue because the result of attribute("add-unique-as") is null if the attribute is not specified, so all database configs except h2 and oracle resulted in NullPointerException.

PR #445, opened 2020-11-01
Replace FileItemWrapper with inline BufferedInputStream.

aabiabdallah:


jonesde:
Have you tested an upload + download round trip with this, specifically with a file uploaded that gets saved in a temporary file?

I haven't tested these changes but from a look at the code my guess is there would be an issue because the InputStream from the original FileItem would have its position advanced by the buffered input stream that wraps it in the executable byte marker check, and then the buffered input stream would have its position reset after reading the first few bytes but the underlying InputStream would NOT have its position reset.

In other words after advancing the FileItem's InputStream it would never get reset so if used directly I would expect that the first X bytes read through the buffered input stream would be lost.

jonesde:
What was the original problem you were trying to solve? You mentioned some things in a mailing list post (which would have been better included here in the pull request for context and easier reference), and it sounds like you want the behavior of the DiskFileItem.getInputStream() method which creates a new InputStream each time it is called. I wasn't aware of that behavior and because of it the issue I described above may not be an issue because the InputStream (buffered or not) for the executable file check would be thrown away and another used the next time it is called. Based on that this code could be simplified even more, closing and throwing away the InputStream used to check for executable byte markers.

The concern I'd have with using that mechanism is that it is not consistent with how most other input streams are handled, especially in the context of HTTP requests where you get one and only one pass through the InputStream. In this case I also wonder about the temporary file delete behavior, ie is it possible it would delete the temporary file after the first close or if it only deletes when the DiskFileItem is destroyed or something.

What is it that you're trying to do where you ran into issues?

aabiabdallah:
True, the input stream is thrown away after the WebUtilities.isExecutable() method completes so no issues there. I've tested this locally and it is working fine. 

The issue we encountered is caused by multiple services accessing the same FileItem in a single HTTP request (flow below). We could very well stick with the currenct implementation but then we will either have to save the FileItem first and force subsequent service calls to work with the saved copy, or we'll have to update each service to mark/reset the input stream in case some other services wishes to access it.

Flow:
1. A validate service: opens the input stream, traverses the file and checks its syntax, then closes the stream. 
2. A generate service: opens the input stream, traverses the file and generates a CSV file, then closes the stream.
3. A save service: saves the file to the DB using the ResourceReference.putStream()

jonesde:
This could be further simplified by not wrapping the InputStream in a BufferedInputStream, and not calling the is.mark(8) method on line 295 in WebUtilities.java (along with removing the reset() call which you already did). Without calls to mark() and reset() the BufferendInputStream is not needed.

Issue #447, opened 2020-11-02
No way to call response.setHeader, instead of addHeader


eigood:
I have integrated keycloak into moqui(more discussion will happen later).  Its KeycloakOIDCFilter sets 'Access-Control-Allow-Credentials: true' in the response.  However, moqui then *also* calls addHeader(), on the same values.  When this happens, google-chrome denies the request, because it sees the duplicate, multi-valued header as 'true, true', and that breaks the invariant it is trying to ensure(namely, it just wants a single header).

I will send a MR for this shortly.

jonesde:
See notes on: https://github.com/moqui/moqui-framework/pull/448

jonesde:
This should be fixed with the changes for #448: https://github.com/moqui/moqui-framework/pull/448

PR #448, opened 2020-11-02
Add a flag to <response-header> that enables calling setHeader instead of addHeader.

eigood:
Fixes: #447


jonesde:
Looks good to me... should also add a boolean type attribute to the xsd file (the functional 'docs'), around here:

https://github.com/moqui/moqui-framework/blob/master/framework/xsd/moqui-conf-2.1.xsd#L262

eigood:
Ack, I'll send an update for the xsd.

jonesde:
I went with a different approach for this, using an 'add' attribute instead of 'single' so that setHeader() is used by default, and addHeader() is optional.

See commit 573d6e6df7f08745362ef0a90886bbb6b44a0f23

PR #451, opened 2020-11-27
Fix the calculation logic error of entity sync time

vrsct:
The current time calculation logic will miss some data, can you check and confirm, thank you

### DIFF
--- a/framework/src/main/groovy/org/moqui/impl/context/ElasticFacadeImpl.groovy
+++ b/framework/src/main/groovy/org/moqui/impl/context/ElasticFacadeImpl.groovy
@@ -695,13 +695,23 @@ class ElasticFacadeImpl implements ElasticFacade {
             if (index == null) return null
             index = index.trim()
             if (index.isEmpty()) return null
-            return indexPrefix != null && !index.startsWith(indexPrefix) ? indexPrefix.concat(index) : index
+            // handle comma separated index names
+            return index.split(",").collect({
+                it = it.trim()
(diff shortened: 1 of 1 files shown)


jonesde:
What was being missed and/or what was the symptom this is trying to address?

The first two changes are in checking/enforcing the delayBufferMillis. Why add the syncSplitMillis to the lastSuccessfulSyncTime? It doesn't make much sense to make the lastSuccessfulSyncTime later in time (adding milliseconds) than it really was.

The changes further down on line 254 subtract syncSplitMillis from the splitFromLong and splitThruLong if splitThruLong is greater than maxSyncLong (which is based on the delayBufferMillis), effectively getting it to iterate more and process time chunks multiple times, or that's what it looks like from a read of the code. What were you trying to do with this change?

vrsct:
The first two changes are realte to line 220 "splitThruLong <= maxSyncLong", the "splitThruLong" is from "splitFromLong + syncSplitMillis".For example,lastSuccessfulSyncTime is 2020:12:05:01,syncSplitMillis is 20000,delayBufferMillis is 60000,current time is 2020:12:06:36,then maxSyncLong is 2020:12:05:36,in a cycle splitThruLong is 2020:12:05:21 and 2020:12:05:41,then the second splitThruLong is bigger than maxSyncLong and the datas between in 2020:12:05:21 and 2020:12:05:41 are not acquired,but the lastSuccessfulSyncTime will be 2020:12:05:41.

PR #458, opened 2021-01-15
Not able to use api_key

dixitdeepak:
Fix truncation error while type casting, Math.round returned incorrect value, and due to this thruDate set as prior date (1970-01-26 02:01:23.647). As per getLoginKey implementation it deletes the old expired login key, as thru date was set as before fromDate it delete the UserLoginKey record immediately.

The changes was one at commit https://github.com/moqui/moqui-framework/commit/4b7c7e70aab19ba5a362ed73cef0fcab65a34f49

### DIFF
--- a/AUTHORS
+++ b/AUTHORS
@@ -101,5 +101,6 @@ Written in 2018-2020 by Ayman Abi Abdallah - aabiabdallah
 Written in 2019 by Daniel Taylor - danieltaylor-nz
 Written in 2020 by Jacob Barnes - Tellan
 Written in 2020 by Amir Anjomshoaa - amiranjom
+Written in 2021 by Deepak Dixit - dixitdeepak
 
 ===========================================================================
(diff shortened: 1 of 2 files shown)


jonesde:
Thank you Deepak, if that fix works that's fine. Looking at the code again I think the error was rounding after adding fromDate.getTime() instead of just rounding the result of the expireHours multiply.

One admin detail: I can merge this only if you sign the AUTHORS file and update this pull request, for more info see:

https://moqui.org/m/docs/moqui/Issue+and+Pull+Request+Guide#intellectual-property

On the other hand this is a small change and I could just do it myself (or an alternative, I'd try first moving the Math.round() to just surround the expireHours multiple and test).

dixitdeepak:
Thanks David for your input, updated code as suggested.
You are right this was due to adding fromDate.getTime in Math.round. 
Moved Math.round to the calculation of expireHours and tested, its working as expected and having more accurate time.

PR #459, opened 2021-01-22
Return menuInclude attribute in getMenuData() so that we can hide or show the subscreen tab in vapps and qapps mode

zhangwei1979:


jonesde:
If I'm guessing correctly this was meant to be used with the changes to hide tabs. A better solution to that is in moqui-runtime commit f103e85 that just only shows the tabs if there is more than 1 tab item. This allows customizing the menu more easily from the server using the existing logic for what is displayed in the header breadcrumb menu, left drawer, and tabs... especially the qvt mode left drawer tree menu which already uses more data from the server for saved finds and will likely use even more in the near future for one level down of subscreen data for all visible screens in the menu (to support expansion to go to a subscreen without going to the parent screen visible in the menu).

PR #463, opened 2021-03-07
fix: apply screen.@transaction-timeout

jenshp:
Fix problem resulting in transaction-timeout in a screen definition never being applied (was always null, thus using the default value), due to 2 separate issues:
1. the String variable with the attribute value is curTxTimeoutAttr, but was attempted to be accessed using the txTimeoutAttr variable name
2. The String to Integer conversion was attempted using the Integer.getInteger() method, which is intended to get a value from the System Properties, not to parse the String which is the intended behavior according to the XSD for this attribute.

### DIFF
--- a/framework/src/main/groovy/org/moqui/impl/screen/ScreenUrlInfo.groovy
+++ b/framework/src/main/groovy/org/moqui/impl/screen/ScreenUrlInfo.groovy
@@ -610,7 +610,7 @@ class ScreenUrlInfo {
             if (curSd.screenNode?.attribute('begin-transaction') == "true") this.beginTransaction = true
             String curTxTimeoutAttr = curSd.screenNode?.attribute("transaction-timeout")
             if (curTxTimeoutAttr) {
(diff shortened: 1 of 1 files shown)


jenshp:
Maybe it is not a bad idea to use Integer.getInteger as a fallback, or even as the primary way, but then we would have to update the XSD comments so the settings can be defined at runtime.

jonesde:
Interesting... IMO this is just a straight up bug in using getInteger() instead of parseInt(). I made simple changes for this in commit 3fda7599. I don't think using getInteger() for Java system properties is all that useful, Moqui has other mechanisms for that which support env vars and Java system properties along with default values defined in the conf xml file, etc.

jenshp:
Great. The commit leaves out the first issue (the String variable with the attribute value is curTxTimeoutAttr, but was attempted to be accessed using the txTimeoutAttr variable name), I will create a new pull request for that one.

jonesde:
Thanks for catching that Jens, hopefully all actually working now...

PR #465, opened 2021-03-11
fix: in ScreenForm.groovy, localize static option.@text that does not need expansion

jenshp:
There was a case left out in localizing options, and it is when a static option is added that does not need expansion for its option.@text attribute.
Not sure whether this solution is the preferred one for performance reasons (expand if the attribute contains the '${' substring, just localize otherwise), or just plainly expand regardless of the '${' substring, which will also localize.


jonesde:
Agreed, it should do this. Because the ResourceFacade.expand() method does l10n and then looks for '${' to see if it needs to actually expand, so the same effect could be had by simply removing "&& text.contains('${')" from the if expression on that original line.

jenshp:
OK, will close this MR and create a new one with that solution

PR #469, opened 2021-03-15
Add list of provinces in Vietnam

donhuvy:


jonesde:
The modified files (CommonL10nData.xml and CurrencyData.xml) do not include a list of Vietnam provinces, is there maybe a file missing?

Also for the currency, would it be more correct to "Viet Nam Dong" as is in this PR, or "Vietnam Dong" with a single word country name?

One last thing, for this contribution please also sign the AUTHORS file (both sections, for the CC0 copyright disclaimer, and the patent disclaimer at the bottom). For more info about that please see:

https://moqui.org/m/docs/moqui/Issue+and+Pull+Request+Guide#intellectual-property

PR #471, opened 2021-04-16
Added support for search-form-inputs in the entity-find-count XML action

aabiabdallah:


zhangwei1979:
This is a runtime error if you open http://localhost:8080/qapps/marble/Party/EditParty?partyId=ORG_ACME

jonesde:
Thank you for reporting that issue. It is fixed in moqui-framework commit 41cfda5e.

My apologies for this, the whitespace trimming was based on my feedback to Ayman and while I ran tests after merging last night I stepped away just after starting the tests so didn't notice until this morning that the change resulted in some errors.

PR #480, opened 2021-06-17
Added support for new no-proxy-forwarding argument in MoquiStart

aabiabdallah:
This addresses potential security issue if Moqui is not running behind a proxy.


jonesde:
To avoid client X-Forwarded-For spoofing (for IP address match like /status and other IP usage like on Visit for auditing purposes, ie tracking for credit card transactions, etc), the first server to handle a request needs to remove existing X-Forwarded-* headers (or set them instead of adding to them).

If there is no proxy running in front of Moqui with Jetty embedded, this change would be needed, and we would need another change to use the same configuration settings and removing checking for X-Forwarded-For in UserFacadeImpl and elsewhere.

If there is a proxy running in front of Moqui with Jetty embedded, then that proxy server should handle the X-Forwarded-* HTTP request headers. There is an issue with some, like AWS ELB & ALB, where if they find an X-Forwarded-For header then it will just append the client IP address to the comma-separated list in the header. The first one is meant to be the original browser (or other client) IP address, but in the case of a spoofed X-Forwarded-For header that would not be the case, it would be the spoofed value and the next value would be the real value.

The best option is for the first server (load balancer or http reverse proxy) to filter out X-Forwarded-* headers, and then add them as a http proxy should.

If that is not possible, we need some sort of configuration to tell Moqui that there are X proxies in front of it, so it should expect X entries in X-Forwarded-For and if there are more than that it should remove from the beginning of the list until the list contains only X entries. This would have to be done in UserFacadeImpl and other Moqui code that uses the X-Forwarded-For header directly, and something would be needed to plug into the embedded Jetty server to do the same (an alternative to ForwardedRequestCustomizer with this logic so that HttpServletRequest.getRemoteAddr() returns the desired value). For other ways of hosting, ie moqui-plus-runtime.war dropped into a servlet container instead of using embedded Jetty, that would have to be handled however the servlet container can handle it (if it can, this is likely not a common solution to X-Forwarded-For spoofing as it is a mess).

PR #488, opened 2021-07-06
Add Second Factor Authentication project Patch

acetousk:
add missing files from the original project


acetousk:
This PR has been tested by me and I don't see anything wrong with it.

PR #495, opened 2021-09-03
Move to JUnit 5 and Gradle 7

MohammadAlHajj:
Dear Mr. Jones
 please find the connected commits to the below repos:
- mantle-usl
- moqui-framework
- moqu-runtime

the goal of these commits is to move to Gradle 7.2 and to Junit 5

gradle 7:
changes need some

-  more testing since i don't know if any more dependencies need to be changed to "java-library"'s 'api' dependency type or kept at gradle's built-in implementation type. 
- These changes rely on the user's projects and which dependencies need their internals  exposed to the user's projects. 
- I also solved the known issue (noted in runtime > base-component > webroot > gradel.build):
  https://github.com/eriwen/gradle-js-plugin/issues/168
  by doing:
  https://github.com/eriwen/gradle-js-plugin/issues/177#issuecomment-705232666
- NOTE: the above project is probably dead and i highly advise the maintainer with the sufficient knowledge to find a replacement.... storing jars in the repo is obviously not an acceptable permanent solution
- Moving to gradle 7 allows the use of the newest Kotlin compiler in our projects and opens the door for moving to Java 11 or possibly java 17 whenit is released 

JUnit5 

- changes are none-intrusive since applying the new Jupiter and Vintage engines together allows both JUnit 4 and 5 tests to be run simultaneously.....  It also separates the result of junit 4 tests into more readable chunks in Intellij (on running "gradle test"). Each test is now part of a subtree with the containing class as the new root of the test group
- Please note that i changed " ignoreFailures = true" since moqui (framework and usl) already has failing tests (on my project) and this blocks the execution of any further tests found in the user's module

For any modification to the commit or for further information, please contact me on Mohammad_Al-Hajj@hotmail.com

### DIFF
--- a/framework/build.gradle
+++ b/framework/build.gradle
@@ -12,19 +12,8 @@
  * <http://creativecommons.org/publicdomain/zero/1.0/>.
  */
 
-tasks.withType(JavaCompile) { options.compilerArgs << "-Xlint:unchecked" }
-tasks.withType(JavaCompile) { options.compilerArgs << "-Xlint:deprecation" }
-tasks.withType(GroovyCompile) { options.compilerArgs << "-Xlint:unchecked" }
-tasks.withType(GroovyCompile) { options.compilerArgs << "-Xlint:deprecation" }
 
-version = '3.0.0-rc8'
 
-apply plugin: 'groovy'
-apply plugin: 'war'
-// to run gradle-versions-plugin use "gradle dependencyUpdates"
-apply plugin: 'com.github.ben-manes.versions'
-// uncomment to add the Error Prone compiler; not enabled by default (doesn't work on Travis CI)
-// apply plugin: 'net.ltgt.errorprone'
 buildscript {
     repositories {
         mavenCentral()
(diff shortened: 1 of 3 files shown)


jonesde:
Thank you Mohammad, this will be helpful. A few weeks ago I was working on updating to Groovy 3, which requires an update to Spock to a version that requires JUnit 5, and I was having trouble getting those working before running out of time, so this should be helpful for the Groovy 3 update effort as well.

On the time topic, my time is pretty limited for the next few weeks so it may be some time before I work on this. I plan to do these sorts of changes as part of the 'java11' branch of moqui-framework so that we can get a few of these big version updates in at the same time, especially because a few are not backward compatible.

In the webroot component build.gradle in moqui-runtime IMO we need to find an alternative sooner than later, and I'd prefer that over another jar variant that we maintain (we already do that for Bitronix), even if it means that we only do simple JS and CSS file combining (just concatenate) and live without the minify.

MohammadAlHajj:
Thank you for the reply David. The next major LTS version of Java (17) will be released on [2021/09/14](https://openjdk.java.net/projects/jdk/17/). Might i suggest that we wait and directly upgrade to that instead of Java 11.

Regarding the webroot gradle js and css minify, here are some solutions that the users of the eriwin project migrated to:

- [gradle-minify-plugin](https://github.com/gradle-webtools/gradle-minify-plugin)
- [terser through npm](https://www.npmjs.com/package/terser)
- [this solution that might not be a good idea as per the comment](https://github.com/eriwen/gradle-js-plugin/issues/177#issuecomment-705445659)

these solutions are not mine nor do i have enough web knowledge to compare their efficiency

jonesde:
It will be a long time after Java 17 is released before we require it in Moqui. Just like Java 11 now, the intent will be to support it for some time while waiting for libraries and infrastructure to catch up (including operating systems, hosting service default images, etc). Part of the community feedback on this includes things like versions of Java supported on all sorts of platforms from AWS Elastic Beanstalk and Container Service to IBM iSeries.

The main impact of this is we can't use new language features until the newer version of Java is required for Moqui, and until fairly recently there were reasons we could not update to Java 11 without breaking things for some Moqui users who told us about that in community feedback.

jonesde:
Nope, no updates, and this could require a fair amount of work. As with most things on my mile long to do list (most of which I'm not interested in), this will get done when I care about it enough AND have sufficient free time and energy for it, or when a client pays to prioritize.

This PR should not go in as-is and I couldn't give an ETA. At some point updating Java, Groovy, Spock, Gradle, etc will become a priority for some reason, but for now other features and such are higher priorities.

MohammadAlHajj:
No worries.... I can work on this myself if you wish, but my pace will be probably the same as yours.

Out of subject, but how can I ask questions like the ones on the mailing list. I currently have a stack of questions/helper tools/feedback/(recommendations?!) i would like to share.

jonesde:
> Out of subject, but how can I ask questions like the ones on the mailing list. I currently have a stack of questions/helper tools/feedback/(recommendations?!) i would like to share.

The mailing list is a fine place, the new forums even better (forum.moqui.org).

Issue #499, opened 2021-11-15
getMenuData() did not return the correct menu name


cyp19901124:
I added some menus to the subscreen_item table.
like this:
![image](https://user-images.githubusercontent.com/13185001/141733702-6c3ebcac-7aed-4618-88e8-7b95bf45e1d6.png)

It can be displayed normally on the interface, but the name changes after clicking.
This is when the menu is not clicked, like this:
![image](https://user-images.githubusercontent.com/13185001/141733860-7b400dfd-8926-41d2-95cd-e14616dcec18.png)

Note the menu example 1:
![image](https://user-images.githubusercontent.com/13185001/141734026-a4265b55-f3b7-44e3-97c8-68cf20b28eb4.png)

The menu name changed after clicking,
I read the code carefully and found that the problem should appear here.
![image](https://user-images.githubusercontent.com/13185001/141734515-bb26f3ca-406e-4902-8665-d2c974148881.png)

My modification suggestions: 
 `String lastTitle = fullUrlInfo.targetScreen.getDefaultMenuName()` 
 Change to
`String lastTitle = curScreen.getSubscreensItem(lastPathItem).menuTitle?:fullUrlInfo.targetScreen.getDefaultMenuName()`

PR #500, opened 2021-11-15
Update ScreenRenderImpl.groovy

cyp19901124:
https://github.com/moqui/moqui-framework/issues/499
Select menuititle for the menu name


jonesde:
Thank you for the idea and patch in this PR. I added a small variation on this that also includes the screens along the path up to the last/target screen in the path, this is in commit 1ebb337.

PR #511, opened 2022-01-22
Fixed the problem that moqui cannot be deployed as non-root webapp in Tomcat

zhangwei1979:


zhangwei1979:
There is a bug is this PR, close it.

PR #512, opened 2022-01-25
Fixed the problem that moqui cannot be deployed as non-root webapp in Tomcat

zhangwei1979:


jonesde:
Thank you Zhang Wei, these changes look fine. I haven't tested thoroughly but will keep an eye out.

PR #516, opened 2022-02-08
Groovy Shell WebSocket Endpoing

jonesde:


chunlinyao:
This PR introduced javax.swing.Timer dependency. I think it maybe better to replaced by java.util.Timer or some self implemented WatchDog class.

PR #526, opened 2022-04-20
Duplicate strategy

jenshp:
Avoid halting because of duplicate jar files, as discussed in https://forum.moqui.org/t/warning-coming-soon-java-11-and-junit-5/219/7?u=jenshp


jonesde:
Thank you Jens, this is helpful. This PR is for the master branch instead of the java11 branch, but it wasn't difficult to find your changes in the build.gradle file. I made some similar changes in commit 1ebc3e3b2359a90b0c19f8f7af2399e832ecb5ed.

Issue #505, opened 2021-12-23
Add a new field type 'json' for service input parameters to make the REST interface simpler.


hansbak:
In order to save some boilerplate, it would be nice to able to recognize Json data from the REST interface.

current code:
```
    <service verb="update" noun="Product">
        <in-parameters>
            <parameter name="product" required="true" type="String"/>
        </in-parameters>
        <actions>
            <script> // convert JSON data
                Object productIn = new groovy.json.JsonSlurper().parseText(product)</script>
          <log message=" incoming json: $productIn "/>
      </actions></service
```
Improvement:

```
    <service verb="update" noun="Product">
        <in-parameters>
            <parameter name="product" required="true" type="Json"/>
        </in-parameters>
        <actions>
          <log message=" incoming decoded json: $product"/>
      </actions></service

```

jonesde:
Quick thought on this: the type attribute specifies the target type, the type the service expects or will convert the specified value to (if possible). A better approach would be to use type="Map" and add a String to Map conversion that assumes the String is in JSON format (and if parsing fails then fail as it does now).

hansbak:
i am creating API documentation with swagger and this becomes more important because the json String I cannot further specify as is possible in a Map.
Can you give a pointer in the code where i should integrate the json slurper?

PR #529, opened 2022-05-11
my proposal for implementation for json input

hansbak:
ok, david my solution to issue https://github.com/moqui/moqui-framework/issues/505 

let me know what you think,

Regards,
Hans


jonesde:
This looks good Hans, I'll marge it and make one small change to use the same JSON parsing as WebFacadeImpl uses (using Jackson with plugins instead of Groovy JsonSlurper).

jonesde:
I made the small change to use the Jackson JSON parser in this commit: ef0c712b

PR #532, opened 2022-07-09
Fix incorrect pathWithParams for savedFinds.

chunlinyao:
Must include parameters for screen.

### DIFF
--- a/framework/src/main/groovy/org/moqui/impl/screen/ScreenRenderImpl.groovy
+++ b/framework/src/main/groovy/org/moqui/impl/screen/ScreenRenderImpl.groovy
@@ -2385,7 +2385,7 @@ class ScreenRenderImpl implements ScreenRender {
                 for (Map<String, Object> userFlf in userFlfList) {
                     EntityValue formListFind = (EntityValue) userFlf.formListFind
                     Map itemMap = [name:formListFind.formListFindId, title:formListFind.description, image:lastImage, imageType:lastImageType,
-                            path:lastPath, pathWithParams:(lastPath + "?formListFindId=" + formListFind.formListFindId)]
+                            path:lastPath, pathWithParams:(currentPath.toString() + (paramString.length() > 0 ? '&' : '?') + "formListFindId=" + formListFind.formListFindId)]
                     if (formListFindId != null && formListFindId.equals(formListFind.formListFindId)) itemMap.active = true
                     savedFindsList.add(itemMap)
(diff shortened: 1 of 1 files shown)


framework/src/main/groovy/org/moqui/impl/screen/ScreenRenderImpl.groovy
```diff
@@ -2385,7 +2385,7 @@ class ScreenRenderImpl implements ScreenRender {
                 for (Map<String, Object> userFlf in userFlfList) {
                     EntityValue formListFind = (EntityValue) userFlf.formListFind
                     Map itemMap = [name:formListFind.formListFindId, title:formListFind.description, image:lastImage, imageType:lastImageType,
-                            path:lastPath, pathWithParams:(lastPath + "?formListFindId=" + formListFind.formListFindId)]
+                            path:lastPath, pathWithParams:(currentPath.toString() + (paramString.length() > 0 ? '&' : '?') + "formListFindId=" + formListFind.formListFindId)]
```
jonesde, line 2388: Thanks, I think I see what was broken here, in cases with parameters resulting in 2 question mark characters in the pathWithParams.

Small detail, and partly messy due to naming of things above, but it would be helpful to be consistent and use either `lastPath` or `currentPath.toString()` in both **path** and **pathWithParams** instead of mixed usage. Otherwise the changes look fine, I suppose another way to fix that would be to look for a `?` in lastPath.

jonesde (changes requested):

framework/src/main/groovy/org/moqui/impl/screen/ScreenRenderImpl.groovy
chunlinyao: And `lastPath` is defined before add parameters to `currentPath`, it doesn't include parameters.

chunlinyao: I cannot found a good solution for this. LINE:2399 below this also use `lastPath` for `path` without parameters, and `currentPath.toString` for `pathWithParams`.
Maybe a utility class can concate query string will improve readability.

chunlinyao: This PR still have a bug. If `paramString` already contains a `formListFindId`, We should remove it first.

PR #533, opened 2022-07-30
On grgit clone, checkout the branch specified in an addons.xml file

acetousk:
Expected behavior:

When cloning a repo with a branch specified in an `addons.xml` file, checkout that branch rather than the default HEAD branch. 

This does the behavior stated above by passing a parameter to checkout the specified branch.

Testing / Edge Cases:
- When no branch is passed, it defaults to checking out the HEAD branch.
- If the branch specified doesn't exist then an empty repository is created instead of defaulting to the HEAD branch.

Functionality changes:
If a branch is specified that doesn't exist before this, it would just default to the HEAD branch. However, after it will result in an empty directory

Note:
We're currently running java 11 and gradle v7.4.2 and [according to the release](https://github.com/ajoberstar/grgit/releases/tag/2.2.1) it doesn't support java 11 or gradle v7.4.2. We should probably upgrade to [grgit 4.1.1](https://plugins.gradle.org/plugin/org.ajoberstar.grgit/4.1.1) or [5.0.0](https://github.com/ajoberstar/grgit/releases/tag/5.0.0). What do you think?

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -980,6 +980,7 @@ File getComponent(String compName, String type, Node addons, Node myaddons, Set 
 }
 File downloadComponent(String targetDirPath, String type, Node component, Node addons, Node myaddons) {
     String compName = component.'@name'
+    String branch = component.'@branch'
 
     String repositoryName = (component.'@repository' ?: myaddons?.'@default-repository' ?: addons.'@default-repository' ?: 'github')
     Node repository = myaddons != null ? (Node) myaddons.repository.find({ it."@name" == repositoryName }) : null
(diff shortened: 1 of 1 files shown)


jonesde:
Looks good, thank you, a branch option is a nice addition.

I was confused about what you said under the "Functionality changes:" line... is that how it works or something you are proposing, or something else? Is the 'after' there about after merging this change commit, or something else?

It would likely be helpful to update grgit, even if it does work fine with Java 11 and Gradle 7.4.2. On a little side note... that link shows the Java versions it was tested with, and while there are some semi-rare exceptions it is usually the case that anything that works on an older version of Java will work on a newer one (ie Java has been mostly backward compatible over time). So far it's working fine, so even if it was made long before Java 11 or Gradle 7 it seems okay.

Anyway, if you're interested in updating the grgit version and testing the main operations (clone with getComponent, gitStatusAll, gitPullAll, gitCheckoutAll) that would be great... because even if it 'seems' to be working fine it may have security updates and performance or other improvements.

Between 5.0.0 and 4.1.1, it looks like there is a potential issue with multi-project builds in 5.0.0 so it might be worth trying but if there are issues better to drop back to 4.1.1 than put in a lot of time getting 5.0.0 to work (I usually trust major version bumps, but sometimes patch releases have fewer issues so it's a good thing to try sooner than later).

acetousk:
> I was confused about what you said under the "Functionality changes:" line... is that how it works or something you are proposing, or something else? Is the 'after' there about after merging this change commit, or something else?

This basically means that if you enter a branch that doesn't exist rather than just pulling the default branch, this code will (for some reason) create an empty directory instead of throwing an error.



> Anyway, if you're interested in updating the grgit version and testing the main operations (clone with getComponent, gitStatusAll, gitPullAll, gitCheckoutAll) that would be great... because even if it 'seems' to be working fine it may have security updates and performance or other improvements.

I've got time. Might as well give it a shot.

PR #538, opened 2022-08-02
Added missing group attribute in entity-definition-3.xsd for view entity

dixitdeepak:


jonesde:
Could you share your thoughts behind this change? To make sure I just checked the code, and this attribute is not used because the group for a view-entity is determined automatically based on the first member-entity (and all member entities must be in the same group of course, doing joins across databases would be a challenge at best).

dixitdeepak:
I am working with data migration effort and using the OFBIZ data model,
while creating view entity by default its using the default group, after adding the group in view-entity its working fine. 

Please let me know if I missed anything?

jonesde:
It sounds like we'll have to get into more detail to make progress on this...

Which version of moqui-framework are you using? (commit ID or something)

What are steps to reproduce, preferably using OOTB entities using small changes to groups, etc?

The idea that only adding this to the XSD would fix the problem seems unlikely. The XSD files are not used when parsing the XML files or at any time when Moqui is running. They are only used by IDEs and such for XML file validation and auto-complete.

dixitdeepak:
Thanks for the details, 
the code working fine, only xsd definition missing, due to missing xsd IDE's won't tell that you can set group as well while defining view entity

gurveenbagga:
I faced a similar issue, the use case was to define a new view entity with member entities from the datasource other than the default one.
 
When we did not use the group attribute in the view definition, it tried to fetch values from the entities from default group name (transactional) as defined in the Moqui Conf file which was not the desired result for us.
 
Upon debugging of the code and analysis of EntityDefinition.groovy file, we figured that the code supports to read from the default group name if attribute is not specified.
https://github.com/hotwax/moqui-framework/blob/master/framework/src/main/groovy/org/moqui/impl/entity/EntityDefinition.groovy#L112
 
So by adding the group attribute to the view entity, it recognised the datasource to make connection with, and returned correct results.

I hope this makes sense, am I missing something?

jonesde:
The issue with the system wide default entity group being used instead of the group from the first member-entity should be fixed in commit 6c1e409b.

Following that change it won't be necessary to use the view-entity.@group attribute to fix the issue. That is still supported, but I don't really want it to be in the entity definition because it is not meant to be used directly.

Along with this change I added back the warning for when a view-entity has member-entity children in more than 1 group. That is related to why I don't like the idea of the use of the view-entity.@group attribute, it should be based on member-entity values and not specified manually.

Of course, when there's a bug in the config loading and defaulting, well... that sort of ruins the whole point... so with that fixed hopefully you won't have to bother with this any more.

PR #540, opened 2022-08-03
Add dockerfile using graalvm

sycured:
[GraalVM](https://www.graalvm.org/java/) doesn't need any code change to obtain better performance

### DIFF
--- a/docker/graalvm/Dockerfile
+++ b/docker/graalvm/Dockerfile
@@ -1,0 +1,43 @@
+# Builds a minimal docker image with openjdk and moqui with various volumes for configuration and persisted data outside the container
+# NOTE: add components, build and if needed load data before building a docker image with this
+
+FROM ghcr.io/graalvm/jdk:ol8-java11
+MAINTAINER Moqui Framework <moqui@googlegroups.com>
+
+WORKDIR /opt/moqui
+
+# for running from the war directly, preffered approach unzips war in advance (see docker-build.sh that does this)
+#COPY moqui.war .
+# copy files from unzipped moqui.war file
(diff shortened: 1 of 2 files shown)


jonesde:
While many deployments have a custom Dockerfile, this default one is used by quite a few and I wouldn't want to make this change without some significant testing.

It might be best to start with a discussion on forum.moqui.org, to get feedback and give others a bit of notice and a chance to comment.

An alternative might be to have a separate Dockerfile for GraalVM. One of the reasons the current Dockerfile is in the /docker/simple directory is so that there can be others, like a docker/graalvm directory or something.

jonesde:
While many deployments have a custom Dockerfile, this default one is used by quite a few and I wouldn't want to make this change without some significant testing.

It might be best to start with a discussion on forum.moqui.org, to get feedback and give others a bit of notice and a chance to comment.

An alternative might be to have a separate Dockerfile for GraalVM. One of the reasons the current Dockerfile is in the /docker/simple directory is so that there can be others, like a docker/graalvm directory or something.

sycured:
Modified to follow the alternative so having docker/graalvm directory

acetousk:
@sycured you can now specify an image for the docker file. See: https://github.com/moqui/moqui-framework/pull/544

jonesde:
That's a good point @acetousk 

Your recent changes to support a build script parameter for the base docker image to use will address this need as well, ie make it easy to use GraalVM. IMO that's a better solution to this need.

What do you think @sycured ?

sycured:
Perfect, it's  easier to use custom base image

PR #541, opened 2022-08-23
Anonymous usage of screens with transitions

jenshp:
A transition defined in a screen with screen.@require-authentication = "anonymous-view" or "anonymous-all" will be considered as not permitted, thus disabling links and forms that attempt to link to that transition.

However, when accessing the same transition directly using plain links, the transition gets executed as expected.

This change avoids the extra check for transitions when the corresponding require-authentication attribute is present, and keeps the check on the service when the transition has a single service call, but honoring the "authenticate" attribute of the service definition considering the authorization level being "view" or "all" according to the screen definition
[updated to reflect changes in last commit for extra service check when transition is single service call.]


acetousk:
@jenshp It looks like there are 4 commits that haven't been merged upstream: https://github.com/moqui/moqui-framework/compare/master...permitTransitionsInAnonymousScreens

Are those 4 commits needed?

PR #546, opened 2022-09-22
Updated slf4j2 impl dependency

dixitdeepak:
Due to a break in compatibility in the SLF4J binding, as of release 2.19.0 two SLF4J to Log4j Adapters are provided
- log4j-slf4j-impl should be used with SLF4J 1.7.x releases or older.
- log4j-slf4j2-impl should be used with SLF4J 2.0.x releases or newer. 

https://logging.staged.apache.org/log4j/2.x/log4j-slf4j-impl/index.html
https://mvnrepository.com/artifact/org.apache.logging.log4j/log4j-slf4j2-impl


jonesde:
Thanks Deepak, good find... when I was looking at this yesterday something seemed off, ie it was odd the release of that one package was missing along with all the other Log4J 2.19.0 releases... so that's where they hid it.

PR #549, opened 2022-10-10
Form list map

acetousk:
Allow a `form-list`'s `list` element to iterate over a map's entries as an entrySet iterator.


jonesde:
Thanks Michael, looks good to me

PR #550, opened 2022-10-30
Add Dockerfile for building and creating an image

acetousk:
As discussed in community meetings. 

Dockerfile builds moqui in a Docker container, then uses the output to create a container suitable for running (copied from docker/simple).

### DIFF
--- a/.dockerignore
+++ b/.dockerignore
@@ -1,0 +1,68 @@
+.git/
+.github/
+.gitlab/
+docker/
+.gitignore
+.gitlab-ci.yml
+.travis.yml
+.whitesource
+AUTHORS
+Dockerfile
+LICENSE.md
+moqui.war
+README.md
+ReleaseNotes.md
+SECURITY.md
+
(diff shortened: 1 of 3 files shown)


acetousk:
Should work, but looking for feedback.

jonesde:
This is an interesting idea, looks like adding a Dockerfile to the root (and a bit more flexible one) instead of replacing the one under docker/simple or adding something like docker/default to put it there.

I'll think on it more, but it might be worth it. It's a bummer adding more files to the root directory, but maybe we should do this and even move the docker directory to a separate repo... I usually use it that way anyway, putting it in a separate repo to configure a particular deployment (if using docker compose files) and it would be nice to just have one to clone. With a moqui-docker-compose separate repo we could also have moqui-kubernetes and other such repos for reference deployment configs.

Thoughts?

acetousk:
I don't think is really needed anymore. I'm closing this pr

PR #551, opened 2022-11-07
Docker image pull feature

Rohit-pawar902:
I have created the Docker-Image-Pull feature in which on clicking INIT button it will fetch the Docker-image mentioned according to credentials and also init Initialise instance in Async ,by refresh you can check status.

- Changes has been verified for : AWS ,Docker-Hub ,Azure Registry
- And Registry-Location : it contains the Registry-server-address now.
- In case of Aws in Password we have to give cmd for generating token and on host system must aws configured by access id and secret key

 EX-:
  Docker-Hub : https://index.docker.io/v1/
  AWS-ECR-registry-server-address :[https://118728153888.dkr.ecr.us-east-1.amazonaws.com/v2/]
  AZURE -REGISTRY -: [rohitmkj.azurecr.io](http://rohitmkj.azurecr.io/)


acetousk:
Thanks for the PR!

It looks like you committed a patch file instead of applying it to your code.

See: https://git-scm.com/docs/git-apply for more information on how to apply a patch.

PR #553, opened 2022-11-08
Docker-Image-Pull Feature

rohitpawar2811:
I have created the Docker-Image-Pull feature in which on clicking the INIT button it will fetch the Docker image mentioned according to credentials and also init Initialise instance in Async,by refresh you can check the status.

- Changes have been verified for : AWS ,Docker-Hub ,Azure Registry
- And Registry-Location: it contains the Registry-server-address now.
- In the case of Aws-Ecr registry in the password we have to give cmd(aws ecr get-login-password --region us-east-1) for generating token and on the host system must aws configured by access id and secret key

**Updated Changes** :
- suggested by @dixitdeepak : Created a field named authTokenCmd into moqui.server.instance.InstanceImage and then we can check based on that if authTokenCmd is not empty then go for that, otherwise use the password for authentication, From this way we can make it more generic and other registries like aws (which generates tokens different each time by using cmd) can easily be configured.

EX-:
- Docker-Hub : https://index.docker.io/v1/
- AWS-ECR-registry-server-address :[https://118728153888.dkr.ecr.us-east-1.amazonaws.com/v2/]
- AZURE -REGISTRY -: [rohitmkj.azurecr.io](http://rohitmkj.azurecr.io/)


Rohit-pawar902:
**Docker-Images-Pulling-Instructions** -: For configuring images ,whose instance we have to create and firstly we have to configure credentials ,from where we have to fetch/pull images.  
-And after following things just create an instance and select the image , which you have created.
**configuration point for image and registry in ** moqui.server.instance.InstanceImage
Instance-Image-Id : Must-be-unique-name/PK your choice
Image-Type-ID -: it's our choice of categorization
Host-Type-Id -: it's our choice of categorization
Image-Name - full-name of image must be given **serveraddress/image-name:tag**
EX-:
- Aws : 118728153888.dkr.ecr.us-east-1.amazonaws.com/tk
- Azure : rohitmkj.azurecr.io/moquil
- Docker-Hub :- rohitroax123/moquil:latest
- Docker-container-Private-registry : localhost:5000/moquil:latest

**Registry-Location : it contains the Registry-server-address now.**

- Docker-container-private-registry : http://localhost:5000/v2/
- Docker-Hub : https://index.docker.io/v1/
- AWS-ECR-registry-server-address :[https://118728153888.dkr.ecr.us-east-1.amazonaws.com/v2/]
- AZURE -REGISTRY -: [rohitmkj.azurecr.io](http://rohitmkj.azurecr.io/)

**Username** : Name of user from which we can login to docker
**Password** : password/ command in case of Aws (through which token is generated)

jonesde:
I apologize for the long delay reviewing this. The changes look good so there is just one thing missing: adding yourself to the AUTHORS file to take more formal credit for your changes, and to grant others clear permission to use them.

There are 2 sections in the AUTHORS file to sign by adding your legal name and github username like the others, and it must be a commit from the named GitHub account. There is one section for the CC0 license and one for the patent disclaimer in addition to the copyright disclaimer for CC0.

framework/service/org/moqui/impl/InstanceServices.xml
```diff
@@ -348,6 +359,42 @@ along with this software (see the LICENSE.md file). If not, see
             ec.logger.info("Initializing docker container for ${appInstance.instanceName} with:\n${jsonBody}")
             restClient.jsonObject(jsonBody)
 
+            try {
+                authToken = instanceImage.password
+                if (((String) instanceImage.username).equals("AWS")) {
```
dixitdeepak, line 364: @rohitpawar2811 
Instead of having check on userName, we can extend InstanceImage entity to store the processCommand

rohitpawar2811: Yes @dixitdeepak you are right I think we can go on this way : 
we can create a field named `authTokenCmd`  into `moqui.server.instance.InstanceImage` and then we can check based on that if authTokenCmd is not empty then go for that , otherwise use password for authentication, From this way we can make it more generic and other registries like aws (which generate token diffrent by using cmd) can easily configured.

rohitpawar2811:
@dixitdeepak  I did the changes which is recommended by , thank you for suggestions

rohitpawar2811:
> I apologize for the long delay reviewing this. The changes look good so there is just one thing missing: adding yourself to the AUTHORS file to take more formal credit for your changes, and to grant others clear permission to use them.
> 
> There are 2 sections in the AUTHORS file to sign by adding your legal name and github username like the others, and it must be a commit from the named GitHub account. There is one section for the CC0 license and one for the patent disclaimer in addition to the copyright disclaimer for CC0.

@jonesde I have added Authors file as you said ,Thank you!

rohitpawar2811:
@jonesde Any thing wrong with my changes?

framework/entity/ServerEntities.xml
```diff
@@ -196,6 +196,7 @@ along with this software (see the LICENSE.md file). If not, see
         <field name="username" type="text-short"/>
         <field name="password" type="text-medium" encrypt="true"/>
         <field name="emailAddress" type="text-medium"/>
+        <field name ="authTokenCmd" type="text-medium"/>
```
acetousk, line 199: Please add a description to `authTokenCmd` and remove the extra space there

framework/service/org/moqui/impl/InstanceServices.xml
```diff
@@ -214,8 +214,13 @@ along with this software (see the LICENSE.md file). If not, see
...
             <set field="hostType" from="appInstance?.instanceHost?.hostType"/>
             <if condition="hostType == null"><return error="true" message="No InstanceHostType found for AppInstance with ID ${appInstanceId}"/></if>
-            <service-call name="${hostType.initService}" in-map="context" out-map="context"/>
-        </actions>
+            <!-- <service-call name="${hostType.initService}" in-map="context" out-map="context"/>-->
+            <!--For async call -->
+            <script><![CDATA[
+            import org.moqui.context.ExecutionContext
+            import java.util.concurrent.Future
+            Future<Map> instanceFuture = ec.service.async().name("${hostType.initService}").parameters([context:context,appInstanceId:appInstanceId]).callFuture()
```
acetousk, line 222: What's the point of having calling this service async?

rohitpawar2811: The point of making this service **async** is to prevent user/client from getting stuck when the image is not available. In synchronous calls, the user would have to wait for the image to be pulled, which could take 6-8 minutes depending on the size. By using asynchronous calls, the user can request to start the container, and if the image is not available, it will be automatically pulled in the background while the user can continue with other tasks. This improves user experience and efficiency.

framework/entity/ServerEntities.xml
rohitpawar2811: Added description to authTokenCmd.
Removed the extra space.
Thanks for pointing it out!

acetousk: You're welcome!

PR #555, opened 2022-11-14
Fixed a runtime error if Currency is BTC

hellozhangwei:


jonesde:
Thank you for reporting this, yes it is a bad assumption that all uomId values are valid Java Currency IDs. This is a fairly small change that I could redo pretty quickly, but if you're up for another round of changes what this needs before going in:

1. this is from an account that isn't in the AUTHORS file (hellozhangwei), are you the same person as 'zhangwei1979' that is in the AUTHORS file? if so maybe easier to submit under that account, if not please sign the AUTHORS file in a commit under this account with a line in each of the 2 sections (top one for CC0 copyright disclaimer, bottom for Apache 2 derived patent disclaimer)
2. small formatting thing, space after 'if' and before the '('
3. instead of logging an error, because this is an expected case better to log a trace message like this one:

```if (logger.isTraceEnabled()) logger.trace("Ignoring NumberFormatException for Timestamp parse: " + e.toString());```

Note that this has an if statement for performance reasons, so the String concatenation is not done if trace is not enabled (ie don't build the argument to a method you know will do nothing with it).

**Comment by hellozhangwei** [comment]
Thanks for your prompt sugguestion.

File: framework/src/main/groovy/org/moqui/impl/context/ElasticFacadeImpl.groovy [hunk 1]
```diff
@@ -196,7 +196,7 @@ class ElasticFacadeImpl implements ElasticFacade {
             requestFactory.init()
 
             // try connecting and get server info
-            int retries = clusterHost == 'localhost' && !"true".equals(System.getProperty("moqui.elasticsearch.started")) ? 1 : 20
+            int retries = ((clusterHost == 'localhost' || clusterHost == '127.0.0.1') && !"true".equals(System.getProperty("moqui.elasticsearch.started"))) ? 1 : 20
```
jonesde, line 199: Just acknowledging this was added, looks fine.

PR #557, opened 2022-11-20
update all code to use docker compose plugin

pythys:
replace all instances of `docker-compose` with `docker compose` as moqui switched to the compose plugin by default

### DIFF
--- a/docker/compose-down.sh
+++ b/docker/compose-down.sh
@@ -8,4 +8,4 @@ fi
 COMP_FILE="${1}"
 
 # set the project name to 'moqui', network will be called 'moqui_default'
-docker-compose -f $COMP_FILE -p moqui down
+docker compose -f $COMP_FILE -p moqui down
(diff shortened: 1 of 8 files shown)


jonesde:
There was some discussion about this recently and a branch with some docker compose updates:

https://forum.moqui.org/t/updating-docker-compose-files-opensearch-compose-plugin-postgres/355

https://github.com/moqui/moqui-framework/tree/dcompose-updates

If you'd like to help with this it would be great, feedback on the forum or with PRs like this to the `dcompose-updates` branch. I think sooner than later we'll need to go from 'docker-compose' to 'docker compose' but I'm not yet sure about the timing so my current plan is to accumulate related changes in a branch and then drop it all in at once with a clear notice about not being backward compatible (due to changes in a dependency in this case).

Anyway, what I'm thinking is better to work from that branch and submit this PR to it rather than the master branch.

PR #558, opened 2022-11-21
switch from docker-compose to docker compose (plugin)

pythys:
as discussed in a [previous PR](https://github.com/moqui/moqui-framework/pull/557) I'm applying the pull request against the proper branch to switch out from `docker-compose` to `docker compose`


pythys:
I just realized I'm updating _from_ master which makes the PR bigger than needed, I will fix and apply another PR

PR #562, opened 2022-12-14
Add relationship called children to EnumerationType

acetousk:
This is probably missing and would be useful for not having to do additional entity finds for the EnumerationType entity

### DIFF
--- a/framework/entity/BasicEntities.xml
+++ b/framework/entity/BasicEntities.xml
@@ -84,6 +84,9 @@ along with this software (see the LICENSE.md file). If not, see
     <entity entity-name="EnumerationType" package="moqui.basic" use="configuration" short-alias="enumerationTypes" cache="true">
         <field name="enumTypeId" type="id" is-pk="true"/>
         <field name="description" type="text-medium"/>
+
+        <relationship type="many" related="moqui.basic.Enumeration" short-alias="children">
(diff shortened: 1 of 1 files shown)


jonesde:
The definition looks good, not sure about the short-alias 'children'. This is following a type many relationship between EnumerationType and Enumeration, ie getting all enum records by type. For type many short-alias values the convention is to use a plural word, singular for type one (and one-nofk). In this case maybe 'enums' would be a more descriptive yet short alias?

When I first saw the PR description with 'children' I was thinking maybe for hierarchical types, but EnumerationType doesn't have that. For what it's worth, when I think of children in a relational data model I usually think of a relationship between records of the same entity, like with parentWorkEffortId and the parent/child relationships with it.

acetousk:
> The definition looks good, not sure about the short-alias 'children'. This is following a type many relationship between EnumerationType and Enumeration, ie getting all enum records by type. For type many short-alias values the convention is to use a plural word, singular for type one (and one-nofk). In this case maybe 'enums' would be a more descriptive yet short alias?
> 
> When I first saw the PR description with 'children' I was thinking maybe for hierarchical types, but EnumerationType doesn't have that. For what it's worth, when I think of children in a relational data model I usually think of a relationship between records of the same entity, like with parentWorkEffortId and the parent/child relationships with it.

I changed `children` to `enums`.

PR #565, opened 2023-02-01
Upgraded freemarker version to 2.3.32, also updated ftl version in FtlTemplateRenderer and MNode class

dixitdeepak:
For more detail
https://freemarker.apache.org/docs/versions_2_3_32.html


jonesde:
Looks good, and thanks for updating the compatibility version constants. This release of FreeMarker should work fine, I tested it during the pre-release and voting.

PR #576, opened 2023-02-23
Handle usernames with different casing in session data

jenshp:
When providing a username with different casing than stored in the DB, login is successful but rendering of next page fails due to not finding the username. This fix makes the search of UserAccount based on the session data to use ignoreCase for the username field.


jonesde:
Looks good, after reviewing the Forum thread it makes sense. Thanks for taking care of this.

PR #578, opened 2023-03-13
Refactor getTitle to prioritize title() method call over data

acetousk:
First use the titleTemplate from the title(String) method, then if the titleText is null or empty use the notification topic's titleTemplate / errorTitleTemplate.

Calling the title method should override the data because the title method is more granular and can change the title on a case by case basis.


jonesde:
Looks good, thank you

PR #583, opened 2023-05-08
Improve handling of service-special when incorrect parameters are given.

eigood:
When using service special, and parameters are given that do not pass validation, moqui will attempt to log this.  However, doing so requires an active transaction.  Service special runs this *after* the current transaction has completed, so the validate logic fails to run, and a NO_TRANSACTION message is printed instead, hiding the root cause of the error.

So, now we implicitly validate the service call in beforeCompletion(), so the current transaction will fail, and the root cause(incorrect parameter specification) can be shown.


acetousk:
Assuming there was some testing on this, this looks good to me. @eigood Have you used this patch in production or a test environment?

PR #584, opened 2023-05-08
Add check-empty-type support to explicit load calls.

eigood:
If "load" is requested from the moqui cli, then all given data load types are *always* run.  However, if moqui is started normally, it has special support to load certain data items only the *first* time, when the database is empty(checked via looking for Enumeration records).

This feature lets one specify check-empty-types, and load types.  Then, always call "load" in the deployment process, and the first time through(empty database) it'll load seed,seed-initial, etc, and subsequent calls will just load seed.

### DIFF
--- a/framework/src/main/java/org/moqui/Moqui.java
+++ b/framework/src/main/java/org/moqui/Moqui.java
@@ -132,10 +132,31 @@ public class Moqui {
 
         // set the data load parameters
         EntityDataLoader edl = ec.getEntity().makeDataLoader();
+        Set<String> loadTypes = new HashSet<>();
         if (argMap.containsKey("types")) {
             String types = argMap.get("types");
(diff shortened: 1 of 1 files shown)


acetousk:
I'm not sure what the objective or use case of this change is. Do you mind explaining that?

It seems like you are adding another way to specify types of data to load conditionally if the database is not loaded (through an enumeration) unless they are specified in the load parameter (where they will always load) in a command line setting.

jonesde:
There is a recently added feature to load data on start, separate from the empty db load, which sounds similar to what you describe.

Here is the commit:

https://github.com/moqui/moqui-framework/commit/9bd047b797a8136f02461997fe3789aac2ba6679

eigood:
I created this feature, then manually added the new lines from the microgram UOM PR, and never bothered to merge this branch back into our own fork.

I shall get back to you on whether 9bd047b works(I suspect it will, given my reading of that change).

PR #585, opened 2023-05-23
Fix regression with partitioned tables in PostgreSQL

chunlinyao:
PostgreSQL JDBC Driver introduced separating type for partitioned table from  40.2.12 pgjdbc/pgjdbc#1708


acetousk:
This looks needed. I'm assuming that this is needed when using PostgreSQL partitioned tables because without this type it will not show in `getTables()` as demonstrated in the PR mentioned.

How will this change affect other databases like MySQL and Oracle (if at all)?

Are there any other changes that are needed to make this work?

chunlinyao:
I haven't tested on MySQL and Oracle, but I think JDBC should ignore it if
it doesn't support.

acetousk ***@***.***> 于 2023年7月23日周日 上午10:30写道：

jonesde:
Hopefully won't cause any problems, I guess we'll find out if it does...

PR #600, opened 2023-06-07
Add subscreensItem.menuInclude to menu data

hellozhangwei:


hellozhangwei:
Think we have following screen structure.

Example.xml
  -- FindExampleAaa.xml (with default-menu-include="true")
  -- FindExampleBbb.xml (with default-menu-include="true")
  -- FindExampleCcc.xml (with default-menu-include="true")
  -- EditExample.xml (with default-menu-include="false")

in this case, if I am on EditExample screen. the EditExample will be included in the menu. I dont want this. So I think we should return subscreensItem.menuInclude in menu data and check subscreensItem.menuInclude in the WebrootVue.qvt.ftl
WebrootVue.qvt.ftl

PR #614, opened 2023-09-12
Currency

jenshp:
Add fields to moqui.basic.Uom to handle currency details, and use them for adding custom currencies or override currencies existing in JVM for both, rounding and formatting currency values in L10nFacade. Also, add a currency-dependent method that hides the symbol but formats according to currency and locale. As discussed in https://forum.moqui.org/t/using-non-iso-4217-currencies/552?u=jenshp

The formatting considers both, the currency and the locale, to determine the correct formatting. This is why the symbol field in the Uom entity is set to be localizable, because some locales use the currency code instead of the symbol, which can be achieved in this implementation by adding a locale-specific symbol for a currency.


jonesde:
Thank you Jens, from a read through it looks good

PR #625, opened 2023-11-17
Implemented withCloseable/try-with-resources where needed in the code…

dixitdeepak:
Initial cleanup of Entity List Iterator for proper resource closure.

This PR addresses occurrences in the codebase where withCloseable/try-with-resources is applied to ensure proper closure of resources. 
Additional attention is needed in xmlActions.groovy.ftl, especially where MiniLang is utilized.

### DIFF
--- a/framework/service/org/moqui/impl/EntitySyncServices.xml
+++ b/framework/service/org/moqui/impl/EntitySyncServices.xml
@@ -236,12 +236,9 @@ along with this software (see the LICENSE.md file). If not, see
                         // ec.logger.warn("=========== get#EntitySyncData entityName=${entryMap.entityName} count=${currentCount} find=${find}")
 
                         if (currentCount > 0) {
-                            EntityListIterator resultEli = find.iterator()
-                            try {
+                            find.iterator().withCloseable ({resultEli ->
                                 int levels = entryMap.dependents ? 2 : 0
                                 resultEli.writeXmlText((Writer) entityWriter, null, levels)
-                            } finally {
-                                resultEli.close()
                             }
                         }
                     }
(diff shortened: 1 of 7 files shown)


jonesde:
For what it's worth, I personally don't like the try with resources code pattern for most things, and where an EntityListIterator is already closed in a finally block I don't see a reason for changing it (and would even prefer not to).

I scanned through and didn't see any, but were there any functional changes in there or places where an ELI was not properly closed?

dixitdeepak:
Thanks David for the review, 
We've encountered memory leaks during the generation of feed files from the database using EntityListIterator (ELI) and subsequent SFTP operations. While we've successfully addressed resource management in the SFTP code by implementing ARM and have observed a reduction in memory leaks, we still face the challenge of server restarts every two weeks to prevent heap memory from reaching 90%.

dixitdeepak:
I am referring the following comment from [ARM](https://docs.oracle.com/javase/tutorial/essential/exceptions/tryResourceClose.html)

`However, this example might have a resource leak. A program has to do more than rely on the garbage collector (GC) to reclaim a resource's memory when it's finished with it. The program must also release the resoure back to the operating system, typically by calling the resource's close method. However, if a program fails to do this before the GC reclaims the resource, then the information needed to release the resource is lost. The resource, which is still considered by the operating system to be in use, has leaked.`

And as per the [The finally Block](https://docs.oracle.com/javase/tutorial/essential/exceptions/finally.html) document
`Important: Use a try-with-resources statement instead of a finally block when closing a file or otherwise recovering resources.`

As per the above comments
If we want to release resources properly, will have to use try with resource for ARM.

PR #629, opened 2024-02-07
Improved log4j2 configuration

AyanF:
In the moqui-framework log4j2.xml configuration file, the Async Appender references the Console Appender and it is positioned before to the Console Appender, while the official log4j2 documentation states that - 

_The AsyncAppender should be configured after the appenders it references to allow it to shut down properly._

Reference - https://logging.apache.org/log4j/log4j-2.3.1/manual/appenders.html#AsyncAppender

Ideally the Async Appender should be placed after all the other Appenders.


jonesde:
Looks good, thank you, merged.

PR #631, opened 2024-02-23
Fixed message queue clearance logic in ec.message.clearAll() method

dixitdeepak:
The ec.message.clearAll() method is intended to clear all messages from the queue, but due to the current implementation, it fails to do so. 

The method operates by first clearing the message list and then calling the clearErrors() method. 
The clearErrors() method is designed to clear error messages from the queue, but as part of its operation, it inadvertently adds an error message back into the messageList. 

This behavior results in the clearAll method being unable to fully clear the message queue, as error messages get reintroduced after the attempt to clear them.


dixitdeepak:
Alternatively we can clear the errorList and validationErrorList list in clearAll(), 
and set the hasErrors = false.

PR #633, opened 2024-05-30
When expiring lock, set lastRunTime to startTime of last successful jobRun (or null)

jenshp:
When a Job Run is interrupted and never ends (e.g. due to a server crash), moqui will at first wait for the expireLockTime specified in the Service Job. After that time, the lock will be cleared, but the lastRunTime would remain at the time the lock was acquired, which is interpreted as the las time the job was successfully run.
This PR fixes this issue, setting the lastRunTime either back to the last successful Job Run, or null if there has not been one. This is the same behavior applied when moqui handles an error during the execution of the Job Run.


acetousk:
I reviewed the code and this looks good to me.

See: https://forum.moqui.org/t/last-run-time-in-scheduledjobrunner/691/3

PR #636, opened 2024-07-11
Add start component to addons and createComponent task to build.grade

acetousk:
Based on https://forum.moqui.org/t/moqui-start-component/725/3


acetousk:
@jonesde Unless you have an objection I'm going to merge this

PR #637, opened 2024-07-15
allow tracking of artifact hits for rest paths (defaults to false)

jenshp:
Changes to enable tracking of artifact hits for rest paths. Together with PR in moqui-runtime repository.


acetousk:
I've looked through this code and it looks great

jenshp:
Added the recommendations talked about today, hope I understood them correctly, let me know if there is some issue.

PR #676, opened 2025-12-16
BIG: upgrade all dependencies of the system

pythys:
- Lots of code cleanups
- centralize gradle dependencyUpdates
- apply new gradle syntax everywhere, especially plugins
- upgrade all libraries to their latest possible versions that don't introduce big breaking changes
- remove old finalize methods
- switch to shiro 2, and away from INI security mananger to programmatic one
- use `threadId()` in code to work interchangeably between virtual and carrier threads
- update all deprecated APIs that crashed or warned due to the upgrade (non exhaustive)
- fix some tests that failed depending on INI file

### DIFF
--- a/framework/service/org/moqui/impl/UserServices.xml
+++ b/framework/service/org/moqui/impl/UserServices.xml
@@ -211,14 +211,14 @@ along with this software (see the LICENSE.md file). If not, see
                     <script>
                         def token = new org.apache.shiro.authc.UsernamePasswordToken((String) userAccount.username, (String) oldPassword)
                         def info = new org.apache.shiro.authc.SimpleAuthenticationInfo(userAccount.username, userAccount.currentPassword,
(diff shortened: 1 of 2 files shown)


framework/src/main/resources/shiro.ini
acetousk: Can we document why we're moving away from the Shiro ini?

pythys: I have an error message when on INI, I will report here

pythys:
Hello! I don't think the constructor works, because whether salted or not, we ALWAYS pass hashed credentials. The constructor you mentioned expects plain, unhashed credentials which don't exist in our database.

I did a round of testing on everything (discovered a few more bugs and made more enhancements) and I can confirm that BOTH salted and unsalted passwords are getting processed correctly. If I remove or change the salt from a salted password they cannot login, and if I add salt to unsalted account then same happens.

But it turns out we have some copy-paste of the code in a few places, and I had to go correct them, because the logic is not isolated to shiro realm class, but also in `UserServices.xml` and I had to fix it over there as well.

PR #706, opened 2026-03-21
Moqui now honors the standard Log4j property 'log4j.configurationFile'

jmochel:
**Intention:** 

To give developers the same level of control during logging of components and the framework they have in other products using Log4j.

**What it does:**

This change  replaces the use of the hard coded `log4j.xml` from the jar resources with 

1. A use of the contents of the standard  Log4J configuration file override property (`log4j.configurationFile`) from system properties or environment variables if it is set in either.
2. If no location is found, it tries to find "log4j2.xml" on the classpath (basically, as a resource bundled with the code or libraries).
3. If it can't find a configuration file in either location, it logs a warning and stops (does not attempt to reconfigure

### DIFF
--- a/framework/src/main/groovy/org/moqui/impl/context/ExecutionContextFactoryImpl.groovy
+++ b/framework/src/main/groovy/org/moqui/impl/context/ExecutionContextFactoryImpl.groovy
@@ -306,14 +306,24 @@ class ExecutionContextFactoryImpl implements ExecutionContextFactory {
     }
 
     protected void reconfigureLog4j() {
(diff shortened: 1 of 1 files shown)


eigood:
It might be nice to see several layers of initialization.  The initial bootstrap, where a *very* simple log4j config is loaded.  Then, after all MoquiConf.xml snippets are loaded and merged, it(log4j) gets re-initialized.

schue:
Typically log4j.configurationFile allows multiple files separated by commas.

If the file doesn't exist it will still try to load the file. A File.exists() that logs a warning if it isn't there would be more defensive.

PR #712, opened 2026-04-10
feat: Optional TimescaleDB hypertables for log tables

pandor4u:
## TimescaleDB Hypertables for Log Tables

Depends on: #700

### Problem
`moqui_logs` and `moqui_http_log` are plain tables that grow indefinitely. The nightly cleanup jobs use row-by-row DELETE which is O(n) and locks the table. No compression or automatic retention.

### Solution
When TimescaleDB is available, automatically:
- **Hypertables**: Convert log tables to hypertables partitioned by `log_timestamp` for efficient chunk-level operations
- **Compression**: Compress chunks older than 7 days (90%+ storage reduction), segmented by `log_level` (logs) and `request_path` (http_log) for efficient filtered queries
- **Retention**: Automatic 90-day retention policy via `add_retention_policy()` — replaces Moqui's nightly DELETE cleanup jobs
- **drop_chunks()**: `deleteByQuery` on log tables uses O(1) chunk drops instead of O(n) per-row deletes

Falls back gracefully to plain tables with BRIN indexes when TimescaleDB is not installed. Safe to run multiple times — skips already-converted hypertables.

### DIFF
--- a/framework/build.gradle
+++ b/framework/build.gradle
@@ -171,6 +171,8 @@ dependencies {
     testImplementation 'org.junit.platform:junit-platform-suite:6.0.1'
     // junit-jupiter-api for using JUnit directly, not generally needed for Spock based tests
     testImplementation 'org.junit.jupiter:junit-jupiter-api:6.0.1'
+    // junit-jupiter-engine required to execute @Test-annotated methods via JUnit Platform
+    testRuntimeOnly 'org.junit.jupiter:junit-jupiter-engine:6.0.1'
     // Spock Framework
     testImplementation platform('org.spockframework:spock-bom:2.4-groovy-5.0') // Apache 2.0
     testImplementation 'org.spockframework:spock-core:2.4-groovy-5.0' // Apache 2.0
@@ -201,6 +203,7 @@ test {
 
     dependsOn cleanTest
     include '**/*MoquiSuite.class'
+    include '**/*PostgresSearchSuite.class'
 
     systemProperty 'moqui.runtime', '../runtime'
     systemProperty 'moqui.conf', 'conf/MoquiDevConf.xml'
--- a/framework/entity/SearchEntities.xml
+++ b/framework/entity/SearchEntities.xml
@@ -1,0 +1,115 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<!--
+This software is in the public domain under CC0 1.0 Universal plus a
+Grant of Patent License.
+
+To the extent possible under law, the author(s) have dedicated all
+copyright and related and neighboring rights to this software to the
+public domain worldwide. This software is distributed without any
+warranty.
+
+You should have received a copy of the CC0 Public Domain Dedication
+along with this software (see the LICENSE.md file). If not, see
+<http://creativecommons.org/publicdomain/zero/1.0/>.
+-->
+<!--
+    PostgreSQL-backed search and logging entities.
+    These tables are created by PostgresElasticClient.initSchema() at startup when type=postgres is configured.
+    Entity definitions here are provided for Moqui entity framework access (queries, etc) — they reference the
+    same underlying tables. The actual CREATE TABLE SQL includes JSONB columns and GIN/BRIN indexes that go
+    beyond what Moqui entities can express, so the DDL is managed by PostgresElasticClient directly.
+-->
+<entities xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+          xsi:noNamespaceSchemaLocation="http://moqui.org/xsd/entity-definition-3.xsd">
+
+    <!-- ======================================================
+         Postgres Search — Document Store (replaces ES indexes)
+         ====================================================== -->
+
+    <!-- moqui_search_index tracks index metadata (equivalent to ES index aliases + mappings storage) -->
+    <!-- authorize-skip="create" allows the internal PostgresElasticClient to create records without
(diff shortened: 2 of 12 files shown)


pandor4u:

This PR adds optional [TimescaleDB](https://www.timescale.com/) hypertable support for the two log tables (`moqui_logs` and `moqui_http_log`), providing automatic compression, retention policies, and O(1) chunk-based cleanup.

### Why this change?

PR #700 stores application logs and HTTP request logs in plain PostgreSQL tables. The production problem:

1. **Unbounded growth**: Log tables grow indefinitely. A moderately busy Moqui instance generates millions of log rows per month.
2. **Expensive cleanup**: Moqui's nightly `CleanupLog` job runs `DELETE FROM moqui_logs WHERE log_timestamp < ?` — this is **O(n)** in the number of matched rows, generates massive WAL (write-ahead log) traffic, bloats the table until VACUUM runs, and holds a lock during execution.
3. **No compression**: Old log data (which is rarely queried) takes the same storage as fresh data.

TimescaleDB solves all three by converting tables to **hypertables** — transparently partitioned by time into "chunks" (typically 7-day intervals). Each chunk is an independent physical table, enabling:
- **`drop_chunks()`**: Deletes an entire chunk in O(1) — a metadata operation, not row-by-row deletion. No WAL bloat, no table lock.
- **Compression**: Old chunks are compressed at 90%+ reduction ratios using columnar encoding.
- **Chunk exclusion**: Time-range queries skip irrelevant chunks entirely (like partition pruning).

### What changed and why

All changes are in **`PostgresElasticClient.groovy`**.

#### Extension detection
```groovy
stmt.execute("CREATE EXTENSION IF NOT EXISTS timescaledb")
hasTimescaleDb = true
```
Same try/catch pattern as other extensions. Falls back cleanly to plain tables.

#### `initTimescaleHypertables(stmt)`
This is the main new method. Called once during `initSchema()` after tables are created.

**Hypertable conversion:**
```sql
SELECT create_hypertable('moqui_logs', 'log_timestamp', migrate_data => true, if_not_exists => true)
```
- `log_timestamp` is the partitioning column — chunks are created per time interval.
- `migrate_data => true` handles the case where the table already has rows (safe for upgrades).
- `if_not_exists => true` makes it idempotent.
- Before calling `create_hypertable`, we check `timescaledb_information.hypertables` to ensure we don't try to convert an already-converted table (the `if_not_exists` flag should handle this, but the explicit check avoids spurious error logs).

**Compression policy:**
```sql
ALTER TABLE moqui_logs SET (timescaledb.compress, timescaledb.compress_segmentby = 'log_level')
SELECT add_compression_policy('moqui_logs', INTERVAL '7 days', if_not_exists => true)
```
- `compress_segmentby = 'log_level'` means each compressed chunk groups rows by log level. This is optimal for queries like "show me all ERROR logs from last week" — the query only decompresses the ERROR segment.
- For `moqui_http_log`, `compress_segmentby = 'request_path'` — optimized for "show me all requests to /api/products".
- 7-day compression threshold balances freshness (recent logs are uncompressed for fast querying) with storage savings.

**Retention policy:**
```sql
SELECT add_retention_policy('moqui_logs', INTERVAL '90 days', if_not_exists => true)
```
TimescaleDB's background worker automatically calls `drop_chunks()` for data older than 90 days. This **replaces** the need for Moqui's nightly `CleanupLog` service job for these tables.

#### `deleteByQueryHttpLog` / `deleteByQueryAppLog` — drop_chunks optimization
When TimescaleDB is available and the delete query has an upper time bound:
```groovy
if (hasTimescaleDb && (range.lte != null || range.lt != null)) {
    return dropChunksForTable("moqui_http_log", range)
}
```
This intercepts Moqui's existing delete-by-query path and routes it through `drop_chunks()` instead of `DELETE`. The existing API contract is preserved — callers don't know the difference.

#### `dropChunksForTable(tableName, range)`
```sql
SELECT drop_chunks('moqui_http_log', older_than => ?)
```
- Returns `-1` because `drop_chunks()` doesn't report a row count (it drops whole chunks).
- **Fallback**: If `drop_chunks()` fails for any reason (e.g., no chunks match, table not a hypertable), falls back to a regular `DELETE FROM ... WHERE log_timestamp <= ?`. This ensures the cleanup always succeeds.

#### Server info
Reports `features: [timescaledb: true/false]` for discoverability.

### What is NOT changed
- **Log table schema is unchanged** — same columns, same indexes. TimescaleDB hypertables are transparent to SQL queries.
- **No new configuration** — activates automatically when TimescaleDB is installed.
- **Non-log tables unaffected** — `moqui_document` and `moqui_search_index` remain plain tables.

### Safety
- `initTimescaleHypertables()` is idempotent — safe to run on every startup.
- `migrate_data => true` handles pre-existing data during conversion.
- All TimescaleDB operations are wrapped in try/catch with fallback behavior.
- The 90-day retention is a guideline — production deployments can adjust via `remove_retention_policy` + `add_retention_policy` with a different interval.

### Performance impact
| Operation | Without TimescaleDB | With TimescaleDB |
|-----------|-------------------|------------------|
| Delete old logs | O(n) row-by-row DELETE | O(1) drop_chunks |
| Storage (90 days logs) | ~100% | ~10-15% (compressed) |
| Time-range query | Full table scan with BRIN | Chunk exclusion + BRIN |
| Insert (current) | Plain INSERT | INSERT (transparent) |

### Depends on
- PR #700 (PostgreSQL search backend, creates `moqui_logs` and `moqui_http_log` tables)

PR #714, opened 2026-04-28
Increase default Docker JVM max heap from 1024m to 4096m

anasomar1:
Production workloads with large data volumes cause OutOfMemoryError crashes
     with the current 1GB max heap default in Docker Compose templates.

     Specific failure scenarios observed in production:
     - Indexing a data document with around 34,000 records and multiple fields
     - Running the scheduled job "clean_ArtifactData_daily" to clean millions of artifact hits was crashing when I restored production data on another server

     Both operations exhausted heap space and crashed the JVM with:
         java.lang.OutOfMemoryError: Java heap space

     Bumping the -Xmx default to 4096m resolved these issues without code changes.

### DIFF
--- a/docker/moqui-acme-postgres.yml
+++ b/docker/moqui-acme-postgres.yml
@@ -87,7 +87,7 @@ services:
       - ./runtime/db:/opt/moqui/runtime/db
       - ./runtime/opensearch:/opt/moqui/runtime/opensearch
     environment:
-      - "JAVA_TOOL_OPTIONS=-Xms1024m -Xmx1024m"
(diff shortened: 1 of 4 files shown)


acetousk:
My main worry here is that this needs to work for systems that have only 1gb allocated to Moqui as that is the effective minimum.

In docker, you can override any environment variable for production environments, and in fact you should configure the enthronement variables to match your production setup.

jonesde:
For production configurations relying on a default max heap is risky, generally always specified unless it is a very basic setup. You are proposing a change to the default value which demonstrates exactly why relying on a default would be risky, and why it is not a great idea for us to change the default without really good reason (like 1GB is no longer adequate, for example; that is not currently the case).

PR #717, opened 2026-06-15
feat: add disable-authz attribute to entity-find/entity-find-one/entity-find-count

jenshp:
Add the same disable-authz attribute of the service-call tag to the entity-find, entity-find-one and entity-find-count tags.

Useful when needing to check business logic (e.g. reservations) where the user normally does not see rows due to EntityFilters, but needs to avoid collissions of resource usage, etc. We had lots of work-arounds in our code switching to groovy-based finds, or creating service-calls for the sole purpose of generating one or two queries with no entity filtering.

### DIFF
--- a/framework/template/XmlActions.groovy.ftl
+++ b/framework/template/XmlActions.groovy.ftl
@@ -122,7 +122,7 @@ ${.node}
 <#macro "entity-find-one">
     <#assign autoFieldMap = .node["@auto-field-map"]?if_exists>
     if (true) {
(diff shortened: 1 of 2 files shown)


dixitdeepak (approved):
This is also useful for SECA rules. If a SECA rule or service job calls another service that has a SECA rule defined and uses entity-find, the system currently blocks the operation because no user is available in the execution context during a scheduled job.

This change would allow the entity-find operation to work correctly for these system-initiated service executions.

PR #742, opened 2026-09-11
Add A2A 1.0 server: JSON-RPC binding, SSE streaming and Agent Card

moqui-industrial:
## Summary

Adds an [A2A 1.0](https://a2a-protocol.org/latest/specification/) server to the `llm-client` branch so that other agents can discover and call Moqui as a remote agent. It sits in front of the existing LLM stack (`LlmClient`, the `assist` profile, SkillIndex) and follows `framework/plans/A2A-Server.md`, whose status section is updated to describe what is implemented, what is deferred and what the proposal superseded.

**A2A is off by default** (`a2a_enabled=false`): the Agent Card and the JSON-RPC endpoint both answer 404 until an operator turns it on, so an upgrade never publishes an agent surface by itself.

## Architecture

```mermaid
flowchart TD
    Servlet["A2AServlet"] --> JsonRpc["A2AJsonRpc"]
    CardServlet["A2ACardServlet"] --> Card["A2ACardBuilderImpl"]
    JsonRpc --> Gateway["A2AGateway"]
    Services["A2AServices.xml"] --> Facade["A2AFacade / A2AFacadeImpl"]
    Components["Moqui components (ec.getA2A())"] --> Facade
    Facade --> Gateway
    Gateway --> Types["A2ATypes"]
    Gateway --> Store["A2ATaskStore"]
    Gateway --> Executor["A2AExecutor / A2AExecutorImpl"]
    Executor --> LLM["LlmFacade"]
    Store --> DB["Entity Engine"]
```

| Layer | Responsibility |
| --- | --- |
| `A2AServlet`, `A2ACardServlet` | HTTP and SSE framing only; no entities, no LLM |
| `A2AJsonRpc` | JSON-RPC envelope, method dispatch, params, error mapping |
| `A2AFacade` (`org.moqui.llm.a2a`), `A2AFacadeImpl` | public API for services and components as `ec.getA2A()`, one instance per `ExecutionContextFactory` |
| `A2AGateway` | binding-neutral operations: context/task resolution, idempotent replay, in-flight rules, blocking or streaming execution |
| `A2ATypes` | protocol semantics: Part shapes, validation, task states, allowed transitions, JSON helpers |
| `A2ATaskStore` | owner-scoped persistence and wire mapping, each operation in a short isolated transaction |
| `A2AExecutor`, `A2AExecutorImpl` | one LLM turn per task: conversation lifecycle, `write_ui` yield to `INPUT_REQUIRED`, resume, cancel, artifact chunks |

Architecture tests keep those boundaries: the servlets and the JSON-RPC layer cannot reach the Entity Engine or the LLM, the store cannot reach the LLM or the transports, and `A2ATypes` stays free of both.

## Protocol surface

- **Persistence** (`moqui.a2a`): contexts, tasks, status history, messages and parts, artifacts (keyed by `taskId` + `artifactId`, since `artifact_id` is unique only within a task), a StreamResponse event log, push configurations (`encrypt="true"`, never returned by the API).
- **JSON-RPC binding**: `POST /llm/a2a/jsonrpc` with the 1.0 method names (`SendMessage`, `SendStreamingMessage`, `GetTask`, `ListTasks`, `CancelTask`, `SubscribeToTask`, `GetExtendedAgentCard`; push methods answer `-32003`), wire params (`id`), raw `Task` results, A2A error codes `-32001`…`-32009` at HTTP 200, `-32700`/`-32600` for parse errors and batches, strict `A2A-Version: 1.0`. Auth is the existing `LlmAuthFilter` on `/llm/*` (session, Basic, `login_key`; permission `LlmGateway`).
- **SSE**: `SendStreamingMessage` streams live on the request thread through `LlmClient.stream` (Task, `WORKING`, `response` artifact chunks from the LLM deltas, final status). `SubscribeToTask` sends the Task and then the persisted events until the task is terminal.
- **Discovery**: `GET /.well-known/agent-card.json` (public, built per request, one generic skill) and `GetExtendedAgentCard` (SkillIndex skills). The advertised URL comes from `a2a_public_url`, or from `X-Forwarded-Proto`/`X-Forwarded-Host` only when `a2a_trust_forwarded_headers` is true: any client can forge those headers.
- **Guarantees**: a task of another user is reported as `TaskNotFoundError`, never as forbidden, and `referenceTaskIds` is checked the same way; terminal states are final; `(userId, messageId)` is unique so a repeated message replays the stored response; per-task `sequenceNum` is allocated under the task row lock with a unique index behind it; a disconnected SSE client never corrupts a task; internal errors are logged and reported generically.
- **Changes to existing files**: `LlmEntities.xml` adds an optional `contextId` to `LlmConversation`; `ExecutionContext`, `ExecutionContextFactory` and their implementations expose `getA2A()`; `framework/build.gradle` adds `**/A2A*Tests.class` to the test includes.

## Deferred

`A2ATaskBus` (live events stay on the request thread, `SubscribeToTask` replays the persisted log, so live subscription across nodes is not supported), the HTTP+JSON/REST binding and gRPC, asynchronous workers and `returnImmediately` (`-32004`), webhook delivery (`-32003`, and it needs SSRF protections first), Bearer/OIDC, a remote `A2AClient`, and `DbResource` storage for `raw` Parts.

Two more, worth doing before this is exposed to untrusted agents:

- **Payload limits.** Neither the number of Parts nor the size of a `raw` Part (base64, stored inline) is capped; only the servlet container limits the request body. Configurable `a2a_max_*` limits rejected with `-32602` would close that.
- **Retention scheduling.** `clean#A2AData` enforces the right rules but no `ServiceJob` runs it, so tasks, events and messages grow until an operator calls it.

One more thing an operator should weigh: the executor gives the model the `request`, `write_ui`, `browse` and `run_service` tools, so a remote agent drives an LLM acting as the authenticated user. Point `a2a_default_profile` at a profile with a reduced tool set and use a least-privilege account rather than an administrator.

## Testing

- `./gradlew :framework:test --tests A2ACoreTests --tests A2AFacadeTests --tests A2AJsonRpcTests --tests LlmServletTests`: 42 + 7 + 10 + 23 pass. They cover protocol validation, ownership across every operation, concurrency (duplicate `messageId`, cancel racing with completion), streaming and disconnects, cleanup rules, the layering, the public facade (single instance, delegation, ownership), and XSD validation of the A2A XML.
- Full suite on a freshly loaded database: 601 tests, the only failures are the ones the base commit already has (`LlmBrowseTests`, `LlmSkillAgentTests`, `LlmSkillTests`, `SecurityAccessControlTests`, `SecurityErrorTests`), which need an LLM provider and search index this environment does not have.
- `framework/test/a2a`: a Hoppscotch collection (14 requests, 37 assertions) and a curl smoke script, both green against a running instance with `a2a_enabled=true`, and 404 on both endpoints with A2A off.
- **Interoperability with the reference implementation**: `framework/test/a2a/interop_a2a_sdk.py` drives the official `a2a-sdk` 1.1.2 client against the server, 13 of 13 checks pass. The client resolves `/.well-known/agent-card.json`, picks the JSONRPC 1.0 interface the card advertises, and then runs SendMessage (task reaches `TASK_STATE_COMPLETED` with the response artifact and both history messages), GetTask, ListTasks, CancelTask, SendStreamingMessage and GetExtendedAgentCard. The stream arrives as the SDK expects: `task`, `status_update`, `artifact_update` chunks, final `status_update`. The SDK also maps our error codes onto its own typed errors, raising `TaskNotCancelableError` on a terminal task and `TaskNotFoundError` on an unknown one.

## Try it

```
a2a_enabled=true ./gradlew load run
curl -s http://localhost:8080/.well-known/agent-card.json
curl -s -u john.doe:moqui -H 'Content-Type: application/json' -H 'A2A-Version: 1.0' \
  -d '{"jsonrpc":"2.0","id":1,"method":"SendMessage","params":{"message":{"messageId":"m1","role":"ROLE_USER","parts":[{"text":"Hello"}]}}}' \
  http://localhost:8080/llm/a2a/jsonrpc
```

`SendMessage` completes when the `assist` profile has a model configured (`llm_openai_url`, `llm_openai_model`, `llm_openai_api_key`); without one the task ends in `TASK_STATE_FAILED`, which the protocol reports correctly.


jonesde:
Found some issues in review but overall this looks good, merging and will follow up in later commits to this llm-client branch. Thanks for the A2A suggestion and for working on this Igor.

